PR #3, opened 2016-02-19
changed includes of logger.h to xlogging.h to avoid redefinition errors

rabbitom:
When compiling with lower version of gcc(3.4.2 in my case), there will be a redefinition error of typedef 'LOGGER_LOG' - the definitions are in logger.h and xlogging.h - and these two files are actually same, so I made this change to avoid the compile error.


dcristoloveanu:
Thanks for the contribution, merging this in.

PR #16, opened 2016-08-24
Fix Memory Leak + Update Samples to Support OpenSSL on Windows

SparhawkSoftware:
Fixes a connection_destroy_endpoint (reported as a bug);

Updates the the samples to use OpenSSL if the use_openssl CMAKE option is set.
Tested against Azure and ActiveMQ.

These changes depend on a matching pull request submitted on the azure-c-shared-utility repo.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "azure-c-shared-utility"]
 	path = azure-c-shared-utility
-	url = https://github.com/Azure/azure-c-shared-utility.git
+	url = https://github.com/SparhawkSoftware/azure-c-shared-utility.git
 [submodule "libwebsockets"]
 	path = libwebsockets
 	url = https://github.com/warmcat/libwebsockets.git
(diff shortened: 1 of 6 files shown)


dcristoloveanu:
@SparhawkSoftware 

Hi Randy,
I started pulling it in, but I would like to keep the existing message_receiver working for EH with minimal changes.

I think it would be best to have 2 samples:
The current message_receiver could be named message_receiver_eh and keep its current content.
You could add the new sample (named message_receiver) that has nothing to do with EventHub and receives from a generic AMQP node and also has the openssl validation callback.

I can also do the changes if you wish on top of what you have.

dcristoloveanu:
If you do the above changes could you please rebase the PR against develop?

SparhawkSoftware:
The samples are not important. You can discard the changes.
I will have my own samples available here: https://github.com/OPCFoundation/UA-AnsiC when I am done.
I need the memory fixes as soon as possible.

dcristoloveanu:
Ok, I started the build to pull in only the first commit in this PR - that makes sure to perform the free and avoid the leak.

dcristoloveanu:
The fix for the memory leak has been merged in develop.

PR #17, opened 2016-09-07
Add overridable install variables for multiarch platforms

srware:
Hardcoding install directories causes problems for multiarch platforms. It's better to use CMake install variables by default and allow users to override for special cases.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -227,8 +227,15 @@ endif()
 set(UAMQP_INC_FOLDER ${CMAKE_CURRENT_LIST_DIR}/inc CACHE INTERNAL "This is the include folder for UAMQP" FORCE)
 set(UAMQP_SRC_FOLDER ${CMAKE_CURRENT_LIST_DIR}/src CACHE INTERNAL "This is the lib folder for UAMQP" FORCE)
 
+# Overridable install variables
+if (CMAKE_INSTALL_LIBDIR)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -227,8 +227,18 @@ endif()
...
+    set (LIB_INSTALL_DIR "lib" CACHE PATH "Library object file directory")
+endif ()
+
 if(WIN32) 
 else() 
-    install (TARGETS uamqp DESTINATION lib) 
-    install (FILES ${uamqp_h_files} DESTINATION include/azureiot/azure_uamqp_c) 
-endif (WIN32) 
+    install (TARGETS uamqp DESTINATION ${LIB_INSTALL_DIR}) 
+    install (FILES ${uamqp_h_files} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/azureiot/azure_uamqp_c)
```
olivierbloch: Why not doing the same thing for include variables?

olivierbloch:
Thanks for your contribution. Defintiively considering merging this in, please take a look at the comment in the code.
Thanks

CMakeLists.txt
srware: If you mean the if catch for CMAKE_INSTALL_LIBDIR I had an issue with an old version of CMake on a particular distro where CMAKE_INSTALL_LIBDIR was not set/unset my the distro package for some reason. I never had this issue with the CMAKE_INSTALL_INCLUDEDIR and to be honest I don't think the catch is necessary with CMake > 2.8 but I left it in as a precaution.

PR #18, opened 2016-09-07
Compile against existing shared util if provided

srware:
Only build the shared util library from scratch if it doesn't already exist. This gives the user the option to compile multiple libraries and SDKs against a common version of the library which can co-exist on the same target.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -144,7 +144,7 @@ endmacro(compileAsC11)
 
 option(memory_trace "set memory_trace to ON if memory usage is to be used, set to OFF to not use it" ON)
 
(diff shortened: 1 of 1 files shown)


olivierbloch:
Thanks for your contribution. We are reviewing it and will very certainly merge it in soon.

tameraw:
@srware - We looked at this change and we're concerned that if there is a desire to rebuild the lib and override the current installed version, this change will not allow for that.
Is building with an already built lib a hard requirement for your project?

andrew-buckley:
Hi @srware, we've completed work which I believe accomplishes what you've done here. Please have a look and close this if that is the case.

Thanks,
-Andrew

dcristoloveanu:
@srware 

Closing this as we believe the work is complete. Feel free to open a new issue/PR (or reopen this one) if something more is needed.

Cheers,
/Dan

PR #28, opened 2017-02-28
Convert message_ut project to use umock_c

dcristoloveanu:
This is one of a series of changes to convert UT projects to using umock_c. The goal is only to convert to umock_c, not also to add/remove tests. Part of this there was a need to remove the usage of amqpalloc and replace it with gballoc.

### DIFF
--- a/tests/message_ut/message_ut.c
+++ b/tests/message_ut/message_ut.c
@@ -1,16 +1,40 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+#ifdef __cplusplus
+#include <cstdlib>
+#include <cstddef>
+#else
+#include <stdlib.h>
+#include <stddef.h>
+#endif
 #include "testrunnerswitcher.h"
-#include "micromock.h"
-#include "micromockcharstararenullterminatedstrings.h"
-#include "azure_uamqp_c/message.h"
-#include "azure_c_shared_utility/xio.h"
-#include "azure_c_shared_utility/socketio.h"
-#include "azure_uamqp_c/frame_codec.h"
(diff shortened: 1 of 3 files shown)


tests/message_ut/message_ut.c
```diff
@@ -114,11 +123,8 @@ TEST_FUNCTION(message_create_succeeds)
 TEST_FUNCTION(message_create_2_times_yields_2_different_message_instances)
```
ewertons: There should be one test to check for general failures, using umock_c_negative_tests_*

tests/message_ut/message_ut.c
```diff
@@ -165,49 +169,46 @@ TEST_FUNCTION(when_allocating_memory_for_the_message_fails_then_message_create_f
 TEST_FUNCTION(message_clone_with_a_valid_argument_succeeds)
```
ewertons: There should be one test to check for general failures, using umock_c_negative_tests_*

tests/message_ut/message_ut.c
dcristoloveanu: Agree, I will add all the tests as a separate PR, I am simply converting now to umock_c.

---
In reply to: [104191365](https://github.com/Azure/azure-uamqp-c/pull/28#discussion_r104191365) [](ancestors = 104191365)

tests/message_ut/message_ut.c
dcristoloveanu: Agree, I will add all the tests as a separate PR, I am simply converting now to umock_c.

---
In reply to: [104191421](https://github.com/Azure/azure-uamqp-c/pull/28#discussion_r104191421) [](ancestors = 104191421)

zolvarga:
:+1:

PR #31, opened 2017-03-03
Fix wrong defaulting of values in amqp_definitions.c

dcristoloveanu:
Fix wrong defaulting of values in amqp_definitions.c


uamqp_generator/amqp_definitions_c.cs
```diff
@@ -19,8 +19,8 @@ namespace amqplib_generator
     /// Class to produce the template output
     /// </summary>
     
-    #line 1 "H:\repos\uamqp\uamqp_generator\amqp_definitions_c.tt"
-    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.VisualStudio.TextTemplating", "12.0.0.0")]
+    #line 1 "D:\repos\azure-uamqp-c\uamqp_generator\amqp_definitions_c.tt"
```
ewertons: Are these hardcoded paths impacting the generated class in any way?

PR #32, opened 2017-03-03
Fixed error in calculating the remote idle time

mario19911:
The value of last_frame_received_time was used uninitialized which
causes a close AMQP frame when an idle timeout was specified.

### DIFF
--- a/src/connection.c
+++ b/src/connection.c
@@ -929,7 +929,7 @@ CONNECTION_HANDLE connection_create2(XIO_HANDLE xio, const char* hostname, const
     }
     else
     {
-        result = (CONNECTION_INSTANCE*)amqpalloc_malloc(sizeof(CONNECTION_INSTANCE));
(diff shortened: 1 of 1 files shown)


dcristoloveanu:
:shipit:

src/connection.c
```diff
@@ -929,7 +929,7 @@ CONNECTION_HANDLE connection_create2(XIO_HANDLE xio, const char* hostname, const
     }
     else
     {
-        result = (CONNECTION_INSTANCE*)amqpalloc_malloc(sizeof(CONNECTION_INSTANCE));
+        result = (CONNECTION_INSTANCE*)amqpalloc_calloc(1, sizeof(CONNECTION_INSTANCE));
```
dcristoloveanu: Interesting.

Actually having a second look at this it seems that we should get the tick counter stamp when creating the connection rather than initialize with 0. Let me have another look at the code.

dcristoloveanu:
After looking at this again, I think the fix should read the tick counter value rather than setting it to 0. I submitted PR #33 for that and I will close this PR.

Thanks for working with us through fixing this.

PR #35, opened 2017-03-06
Convert sasl_plain unit tests to use umock_c

dcristoloveanu:
Convert sasl_plain unit tests to use umock_c instead of the old Micromock C++.
Also replaced usage of amqpalloc with gballoc for the sasl_plain module.

### DIFF
--- a/tests/sasl_plain_ut/sasl_plain_ut.c
+++ b/tests/sasl_plain_ut/sasl_plain_ut.c
@@ -1,62 +1,91 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+#ifdef __cplusplus
 #include <cstdint>
+#include <cstddef>
+#else
+#include <stdint.h>
+#include <stddef.h>
+#endif
 #include "testrunnerswitcher.h"
-#include "micromock.h"
-#include "micromockcharstararenullterminatedstrings.h"
+#include "umock_c.h"
+#include "umocktypes_charptr.h"
+
+void* my_gballoc_malloc(size_t size)
+{
+    return malloc(size);
+}
+
+void my_gballoc_free(void* ptr)
+{
+    free(ptr);
+}
+
+#define ENABLE_MOCKS
+
+#include "azure_c_shared_utility/gballoc.h"
+
+#undef ENABLE_MOCKS
+
 #include "azure_uamqp_c/sasl_plain.h"
 
 #define MAX_AUTHZID "1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890" "1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890" "1234567890123456789012345678901234567890123456789012345"
 #define MAX_AUTHCID MAX_AUTHZID
 #define MAX_PASSWD MAX_AUTHZID
 
-TYPED_MOCK_CLASS(amqp_frame_codec_mocks, CGlobalMock)
-{
-public:
-	/* amqpalloc mocks */
-	MOCK_STATIC_METHOD_1(, void*, amqpalloc_malloc, size_t, size)
-	MOCK_METHOD_END(void*, malloc(size));
-	MOCK_STATIC_METHOD_1(, void, amqpalloc_free, void*, ptr)
-		free(ptr);
-	MOCK_VOID_METHOD_END();
-};
+static TEST_MUTEX_HANDLE g_testByTest;
+static TEST_MUTEX_HANDLE g_dllByDll;
 
-extern "C"
+DEFINE_ENUM_STRINGS(UMOCK_C_ERROR_CODE, UMOCK_C_ERROR_CODE_VALUES)
+
+static void on_umock_c_error(UMOCK_C_ERROR_CODE error_code)
 {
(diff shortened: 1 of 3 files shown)


tests/sasl_plain_ut/sasl_plain_ut.c
```diff
@@ -65,18 +94,18 @@ TEST_FUNCTION_CLEANUP(method_cleanup)
...
+	EXPECTED_CALL(gballoc_malloc(IGNORED_NUM_ARG));
 
 	// act
-	CONCRETE_SASL_MECHANISM_HANDLE result = saslplain_create(&sasl_plain_config);
+	result = saslplain_create(&sasl_plain_config);
 
 	// assert
 	ASSERT_IS_NOT_NULL(result);
-	mocks.AssertActualAndExpectedCalls();
+	ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
```
ewertons: Very good

tests/sasl_plain_ut/sasl_plain_ut.c
```diff
@@ -186,32 +214,30 @@ TEST_FUNCTION(saslplain_create_with_NULL_authzis_succeeds)
...
 
 	// assert
-	// uMock checks the calls
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
 }
 
 /* Tests_SRS_SASL_PLAIN_01_006: [If the argument concrete_sasl_mechanism is NULL, saslplain_destroy shall do nothing.] */
 TEST_FUNCTION(saslplain_destroy_with_NULL_handle_does_nothing)
 {
 	// arrange
```
ewertons: Should add umock_c_reset_all_calls();

tests/sasl_plain_ut/sasl_plain_ut.c
```diff
@@ -86,95 +115,94 @@ TEST_FUNCTION(saslplain_create_with_valid_args_succeeds)
...
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
+    ASSERT_IS_NULL(result);
 }
 
 /* Tests_SRS_SASL_PLAIN_01_004: [If either the authcid or passwd member of the config structure is NULL, then saslplain_create shall fail and return NULL.] */
 TEST_FUNCTION(saslplain_create_with_NULL_passwd_fails)
 {
 	// arrange
-	amqp_frame_codec_mocks mocks;
 	SASL_PLAIN_CONFIG sasl_plain_config = { "test_authcid", NULL, "test_authzid" };
```
ewertons: Should add umock_c_reset_all_calls();

tests/sasl_plain_ut/sasl_plain_ut.c
```diff
@@ -577,17 +593,16 @@ TEST_FUNCTION(saslplain_challenge_with_NULL_response_bytes_fails)
...
 	// act
 	const SASL_MECHANISM_INTERFACE_DESCRIPTION* result = saslplain_get_interface();
 
 	// assert
-	ASSERT_ARE_EQUAL(void_ptr, (void_ptr)saslplain_create, (void_ptr)result->concrete_sasl_mechanism_create);
-	ASSERT_ARE_EQUAL(void_ptr, (void_ptr)saslplain_destroy, (void_ptr)result->concrete_sasl_mechanism_destroy);
-	ASSERT_ARE_EQUAL(void_ptr, (void_ptr)saslplain_get_init_bytes, (void_ptr)result->concrete_sasl_mechanism_get_init_bytes);
-	ASSERT_ARE_EQUAL(void_ptr, (void_ptr)saslplain_get_mechanism_name, (void_ptr)result->concrete_sasl_mechanism_get_mechanism_name);
-	ASSERT_ARE_EQUAL(void_ptr, (void_ptr)saslplain_challenge, (void_ptr)result->concrete_sasl_mechanism_challenge);
+	ASSERT_IS_NOT_NULL(result->concrete_sasl_mechanism_create);
```
ewertons: There is a loss of granularity here. 
Before the contents were checked for expected value, now it's only checking for not NULL.

Would suggesting changing it to check the actual expected pointer values, like before.

tests/sasl_plain_ut/sasl_plain_ut.c
dcristoloveanu: Not needed as it is done before each test function in the test function init.

---
In reply to: [104449908](https://github.com/Azure/azure-uamqp-c/pull/35#discussion_r104449908) [](ancestors = 104449908)

tests/sasl_plain_ut/sasl_plain_ut.c
dcristoloveanu: Not needed as it is done in test function init.

---
In reply to: [104450017](https://github.com/Azure/azure-uamqp-c/pull/35#discussion_r104450017) [](ancestors = 104450017)

tests/sasl_plain_ut/sasl_plain_ut.c
dcristoloveanu: Yes, true, the check is weaker now. I plan to have another change on the tests and spec for these modules implementing an interface so that we directly call the functions like saslplain_get_interface()->challenge() etc. in the tests. But that has to be another change as I would like to keep changes confined so they are reviewable.

---
In reply to: [104450954](https://github.com/Azure/azure-uamqp-c/pull/35#discussion_r104450954) [](ancestors = 104450954)

PR #37, opened 2017-03-08
Remove calling the send complete function when send fails right away

dcristoloveanu:
-Remove calling the send complete function when send fails right away
-Fix a leak in case the message send errors due to unknown body type.


jebrando:
:shipit:

PR #43, opened 2017-03-10
Pick up new websocket client changes from C shared utility

dcristoloveanu:
Pick up new websocket client changes from C shared utility


zolvarga:
:shipit:

dcristoloveanu:
The part where the samples are updated to use the new WSIO_CONFIG has been checked in. The other change (to remove linking the lws lib) will come as separate PR.

PR #44, opened 2017-03-13
Convert saslclientio unittests to umock_c. Also replace usage of amqp…

dcristoloveanu:
Convert saslclientio unittests to umock_c. Also replace usage of amqpallloc with gballoc


zolvarga:
:shipit:

PR #45, opened 2017-03-13
Remove obsolete socketio_win32 uni tests

dcristoloveanu:
Remove obsolete socketio_win32 uni tests - empty anyhow


tests/socketio_win32_ut/CMakeLists.txt
```diff
@@ -1,24 +0,0 @@
-#Copyright (c) Microsoft. All rights reserved.
```
zolvarga: ># [](start = 0, length = 1)

We cannot add comment in the metadata so I write it here the nitpick: Fix in the commit messagethe "uni"....

zolvarga:
:shipit:

tests/socketio_win32_ut/CMakeLists.txt
dcristoloveanu: Thanks, fixing. it.

---
In reply to: [105763370](https://github.com/Azure/azure-uamqp-c/pull/45#discussion_r105763370) [](ancestors = 105763370)

PR #53, opened 2017-03-15
Remove amqpalloc alltogether

dcristoloveanu:
Remove amqpalloc alltogether


dcristoloveanu:
Yep, we killed amqpalloc, since it was duplicating pretty much what gballoc was doing for the only added option of enabling/disabling the memory trace at runtime. If that turns out to be very useful or screamed for, we can add that to gballoc.

PR #63, opened 2017-03-29
Update AMQP management API to align to a proper async semantics (add …

dcristoloveanu:
Update AMQP management API to align to a proper async semantics (add open_complete callback, on_error callback, etc.)

This is in preparation of writing unit tests for AMQP management.

### DIFF
--- a/devdoc/cbs_requirements.md
+++ b/devdoc/cbs_requirements.md
@@ -11,7 +11,7 @@
 	{
 		CBS_OPERATION_RESULT_OK,
 		CBS_OPERATION_RESULT_CBS_ERROR,
-		CBS_OPERATION_RESULT_OPERATION_FAILED
+		AMQP_MANAGEMENT_EXECUTE_OPERATION_FAILED_BAD_STATUS
 	} CBS_OPERATION_RESULT;
 
     typedef enum CBS_OPEN_COMPLETE_RESULT_TAG
@@ -38,7 +38,7 @@
 ### cbs_create
 
 ```c
-MOCKABLE_FUNCTION(, CBS_HANDLE, cbs_create, SESSION_HANDLE, session, ON_AMQP_MANAGEMENT_STATE_CHANGED, on_amqp_management_state_changed, void*, callback_context);
+MOCKABLE_FUNCTION(, CBS_HANDLE, cbs_create, SESSION_HANDLE, session);
 ```
 
 XX**SRS_CBS_01_001: [** `cbs_create` shall create a new CBS instance and on success return a non-NULL handle to it. **]**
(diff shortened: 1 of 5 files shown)


devdoc/cbs_requirements.md
```diff
@@ -38,16 +38,16 @@
 ### cbs_create
 
 ```c
-MOCKABLE_FUNCTION(, CBS_HANDLE, cbs_create, SESSION_HANDLE, session, ON_AMQP_MANAGEMENT_STATE_CHANGED, on_amqp_management_state_changed, void*, callback_context);
+MOCKABLE_FUNCTION(, CBS_HANDLE, cbs_create, SESSION_HANDLE, session);
 ```
 
 XX**SRS_CBS_01_001: [** `cbs_create` shall create a new CBS instance and on success return a non-NULL handle to it. **]**
```
ewertons: What are these XX on the requirements?

inc/azure_uamqp_c/amqp_management.h
```diff
@@ -14,34 +14,34 @@ extern "C" {
...
 
-    typedef enum OPERATION_RESULT_TAG
+    typedef enum AMQP_MANAGEMENT_EXECUTE_OPERATION_RESULT_TAG
     {
-        OPERATION_RESULT_OK,
-        OPERATION_RESULT_ERROR,
-        OPERATION_RESULT_OPERATION_FAILED
-    } OPERATION_RESULT;
+        AMQP_MANAGEMENT_EXECUTE_OPERATION_OK,
+        AMQP_MANAGEMENT_EXECUTE_OPERATION_ERROR,
```
ewertons: What's the difference between ERROR and FAILED? Could you add a comment in front to clarify?

src/amqp_management.c
```diff
@@ -281,14 +279,53 @@ static void on_message_sender_state_changed(void* context, MESSAGE_SENDER_STATE
         break;
 
     case MESSAGE_SENDER_STATE_OPEN:
-        amqp_management_instance->sender_connected = 1;
-        (void)send_operation_messages(amqp_management_instance);
+        switch (amqp_management_instance->amqp_management_state)
+        {
+        default:
+        case AMQP_MANAGEMENT_STATE_ERROR:
+            /* do nothing */
```
ewertons: Should LogError perhaps?

src/amqp_management.c
```diff
@@ -281,14 +279,53 @@ static void on_message_sender_state_changed(void* context, MESSAGE_SENDER_STATE
...
+            break;
+
+        case AMQP_MANAGEMENT_STATE_OPENING:
+            amqp_management_instance->sender_connected = -1;
```
ewertons: ```amqp_management_instance->sender_connected = -1``` doesn't make sense if you are testing it right below.

What's the intent here?

src/amqp_management.c
```diff
@@ -303,14 +340,54 @@ static void on_message_receiver_state_changed(const void* context, MESSAGE_RECEI
...
+        default:
+        case AMQP_MANAGEMENT_STATE_ERROR:
+            /* do nothing */
+            break;
+
+        case AMQP_MANAGEMENT_STATE_OPENING:
+            amqp_management_instance->receiver_connected = -1;
+
+            if ((amqp_management_instance->sender_connected != 0) &&
+                (amqp_management_instance->receiver_connected != 0))
```
ewertons: What's the intent of testing receiver_connected if it was just set above?

devdoc/cbs_requirements.md
dcristoloveanu: XX is implemented and tested.

src/amqp_management.c
dcristoloveanu: I am not re-writing with tests amqp_management.c as part of this change. This will come in a subsequent PR in order to keep things scoped.

src/amqp_management.c
dcristoloveanu: Actually only one should be set will, the sender one for message sender and the receiver one for messagereceiver. I'll fix that.

src/amqp_management.c
dcristoloveanu: Actually only one should be set will, the sender one for message sender and the receiver one for messagereceiver. I'll fix that.

inc/azure_uamqp_c/amqp_management.h
dcristoloveanu: FAILED means that the status code is a failure code. I'll change the name of the enum to reflect that.

dcristoloveanu:
Thanks!

PR #68, opened 2017-03-29
Update readme

dcristoloveanu:
Update readme

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -2,34 +2,36 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
 
 # uAMQP
 
-uAMQP is a general purpose C library for AMQP.
+`uAMQP` is a general purpose C library for AMQP 1.0.
 
 The goal is to be as compliant with the standard as possible while optimizing for low RAM footprint and also being portable.
 
 It is currently a client side implementation only. Although much of the standard is symmetrical, there are parts that are asymmetrical, like the SASL handshake.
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -2,34 +2,36 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
 
 # uAMQP
 
-uAMQP is a general purpose C library for AMQP.
+`uAMQP` is a general purpose C library for AMQP 1.0.
```
ewertons: I never got to ask this... why it's 1.0, but uamqp prints 3.1.0.0 ?

-> Header (AMQP 3.1.0.0)
<- Header (AMQP 3.1.0.0)

From:
https://github.com/Azure/azure-uamqp-c/blob/master/src/saslclientio.c#L159
and
https://github.com/Azure/azure-uamqp-c/blob/master/src/saslclientio.c#L363

The AMQP 1.0. spec mentions:
Prior to sending any Frames on a Connection, each peer MUST start by sending a protocol header
that indicates the protocol version used on the Connection. The protocol header consists of the
upper case ASCII letters "AMQP" followed by a protocol id of zero, followed by three unsigned
bytes representing the major, minor, and revision of the protocol version (currently 1 (MAJOR), 0
(MINOR), 0 (REVISION)).

dcristoloveanu: 3.1.0.0 is the SASL header. It is in chapter 5.3 of the spec.

dcristoloveanu:
Thanks!

PR #74, opened 2017-04-04
Generate random port number for local_client_server_tcp_e2e test

ewertons:
Please review the code changes. Thanks.

### DIFF
--- a/tests/local_client_server_tcp_e2e/local_client_server_tcp_e2e.c
+++ b/tests/local_client_server_tcp_e2e/local_client_server_tcp_e2e.c
@@ -26,6 +26,22 @@ static TEST_MUTEX_HANDLE g_dllByDll;
 
 #define TEST_TIMEOUT 30 // seconds
 
+static int generate_port_number(void)
+{
+    int port_number;
+    const char* comp_time = __TIME__;
+    int comp_secs = 10 * (comp_time[6] - '0') + (comp_time[7] - '0');
+    int comp_mins = 10 * (comp_time[3] - '0') + (comp_time[4] - '0');
+    int comp_hours = 10 * (comp_time[0] - '0') + (comp_time[1] - '0');
+
+    port_number = 5672 + (int)(5000 * (double)rand() / RAND_MAX) + comp_hours + comp_mins + comp_secs; // pseudo random number.
+
+    LogInfo("Generated port number: %d", port_number);
+
+    return port_number;
+}
+
+
 BEGIN_TEST_SUITE(local_client_server_tcp_e2e)
 
 TEST_SUITE_INITIALIZE(suite_init)
@@ -38,6 +54,8 @@ TEST_SUITE_INITIALIZE(suite_init)
 
     result = platform_init();
     ASSERT_ARE_EQUAL_WITH_MSG(int, 0, result, "platform_init failed");
+
+    srand((unsigned int)time(NULL));
 }
 
 TEST_SUITE_CLEANUP(suite_cleanup)
@@ -146,8 +164,9 @@ static void on_socket_accepted(void* context, XIO_HANDLE io)
 TEST_FUNCTION(client_and_server_connect_and_send_one_message_settled)
 {
     // arrange
+    int port_number = generate_port_number();
     SERVER_INSTANCE server_instance;
-    SOCKET_LISTENER_HANDLE socket_listener = socketlistener_create(5672);
+    SOCKET_LISTENER_HANDLE socket_listener = socketlistener_create(port_number);
     int result;
     XIO_HANDLE socket_io;
     CONNECTION_HANDLE client_connection;
@@ -173,7 +192,8 @@ TEST_FUNCTION(client_and_server_connect_and_send_one_message_settled)
     ASSERT_ARE_EQUAL_WITH_MSG(int, 0, result, "socketlistener_start failed");
 
     // start the client
-    SOCKETIO_CONFIG socketio_config = { "localhost", 5672, NULL };
+    SOCKETIO_CONFIG socketio_config = { "localhost", 0, NULL };
+    socketio_config.port = port_number;
     socket_io = xio_create(socketio_get_interface_description(), &socketio_config);
     ASSERT_IS_NOT_NULL_WITH_MSG(socket_io, "Could not create socket IO");
 
@@ -256,8 +276,9 @@ TEST_FUNCTION(client_and_server_connect_and_send_one_message_settled)
 TEST_FUNCTION(client_and_server_connect_and_send_one_message_unsettled)
 {
     // arrange
+    int port_number = generate_port_number();
     SERVER_INSTANCE server_instance;
-    SOCKET_LISTENER_HANDLE socket_listener = socketlistener_create(5672);
+    SOCKET_LISTENER_HANDLE socket_listener = socketlistener_create(port_number);
     int result;
     XIO_HANDLE socket_io;
     CONNECTION_HANDLE client_connection;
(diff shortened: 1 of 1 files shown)


tests/local_client_server_tcp_e2e/local_client_server_tcp_e2e.c
```diff
@@ -26,6 +26,23 @@ static TEST_MUTEX_HANDLE g_dllByDll;
...
+	int port_number;
+	const char* comp_time = __TIME__;
+	int comp_secs = 10 * (comp_time[6] - '0') + (comp_time[7] - '0');
+	int comp_mins = 10 * (comp_time[3] - '0') + (comp_time[4] - '0');
+	int comp_hours = 10 * (comp_time[0] - '0') + (comp_time[1] - '0');
+
+	srand((unsigned int)time(NULL));
+	port_number = 5672 + (int)(5000 * (double)rand() / RAND_MAX) + comp_hours + comp_mins + comp_secs; // pseudo random number.
+
+	LogInfo("Generated port number: %d", port_number);
```
jebrando: >LogInfo("Generated port number: %d", port_number); [](start = 1, length = 50)

Do you want to print this out every time?

tests/local_client_server_tcp_e2e/local_client_server_tcp_e2e.c
```diff
@@ -26,6 +26,23 @@ static TEST_MUTEX_HANDLE g_dllByDll;
 
 #define TEST_TIMEOUT 30 // seconds
 
+static int generate_port_number()
+{
+	int port_number;
+	const char* comp_time = __TIME__;
+	int comp_secs = 10 * (comp_time[6] - '0') + (comp_time[7] - '0');
+	int comp_mins = 10 * (comp_time[3] - '0') + (comp_time[4] - '0');
+	int comp_hours = 10 * (comp_time[0] - '0') + (comp_time[1] - '0');
```
jebrando: this may be unnecessary because the test can start at the same time.

tests/local_client_server_tcp_e2e/local_client_server_tcp_e2e.c
```diff
@@ -26,6 +26,23 @@ static TEST_MUTEX_HANDLE g_dllByDll;
 
 #define TEST_TIMEOUT 30 // seconds
 
+static int generate_port_number()
+{
+	int port_number;
```
jebrando: >	 [](start = 0, length = 1)

Can you have all the indentions as tabs or space for the whole file please

tests/local_client_server_tcp_e2e/local_client_server_tcp_e2e.c
```diff
@@ -256,8 +275,9 @@ TEST_FUNCTION(client_and_server_connect_and_send_one_message_settled)
 TEST_FUNCTION(client_and_server_connect_and_send_one_message_unsettled)
 {
     // arrange
+	int port_number = generate_port_number();
```
jebrando: >	 [](start = 0, length = 1)

Change this to spaces or change all spaces to tabs.  Here and everywhere.

tests/local_client_server_tcp_e2e/local_client_server_tcp_e2e.c
```diff
@@ -284,6 +304,7 @@ TEST_FUNCTION(client_and_server_connect_and_send_one_message_unsettled)
 
     // start the client
     SOCKETIO_CONFIG socketio_config = { "localhost", 5672, NULL };
```
jebrando: >5672 [](start = 53, length = 4)

This Number should be port_number or zero.

tests/local_client_server_tcp_e2e/local_client_server_tcp_e2e.c
```diff
@@ -174,6 +192,7 @@ TEST_FUNCTION(client_and_server_connect_and_send_one_message_settled)
...
```
jebrando: >5672 [](start = 53, length = 4)

port_number here.

jebrando:
:shipit:

tests/local_client_server_tcp_e2e/local_client_server_tcp_e2e.c
```diff
@@ -26,6 +26,23 @@ static TEST_MUTEX_HANDLE g_dllByDll;
 
 #define TEST_TIMEOUT 30 // seconds
 
+static int generate_port_number()
```
dcristoloveanu: >() [](start = 31, length = 2)

void, because it is C

tests/local_client_server_tcp_e2e/local_client_server_tcp_e2e.c
ewertons: Yes, if the test ever fails, at least we can have the number to debug.

tests/local_client_server_tcp_e2e/local_client_server_tcp_e2e.c
```diff
@@ -26,6 +26,23 @@ static TEST_MUTEX_HANDLE g_dllByDll;
...
+{
+	int port_number;
+	const char* comp_time = __TIME__;
```
dcristoloveanu: >__TIME__ [](start = 25, length = 8)

Let's just use rand() as Jelani said. Adding extra salt cannot be quantified, since the __TIME__ will be the same in some cases anyhow. __TIME__ is expanded at preprocessor time anyhow. Are you sure the translation unit even gets compiled more than one time?

Taking all of this into account, I'd say let's go with the simple rand for now. If you want bulletproof I think you need another solution.

tests/local_client_server_tcp_e2e/local_client_server_tcp_e2e.c
dcristoloveanu: Yes, please remove it and leave rand.

---
In reply to: [109758964](https://github.com/Azure/azure-uamqp-c/pull/74#discussion_r109758964) [](ancestors = 109758964)

dcristoloveanu:
:shipit:

tests/local_client_server_tcp_e2e/local_client_server_tcp_e2e.c
ewertons: Actually this is the part that ensures that two instances of the tests don't fail if they start at the same time.
__TIME__ is assigned during compilation; since our exe, _valgrind and _helgrind builds don't happen in parallel, this is the most distinctive info we can use to avoid collisions

tests/local_client_server_tcp_e2e/local_client_server_tcp_e2e.c
ewertons: Sure

tests/local_client_server_tcp_e2e/local_client_server_tcp_e2e.c
ewertons: Sure

tests/local_client_server_tcp_e2e/local_client_server_tcp_e2e.c
ewertons: Ok. I can't assign port_number because this initializer only takes constant values.

tests/local_client_server_tcp_e2e/local_client_server_tcp_e2e.c
ewertons: Ok. I can't assign port_number because this initializer only takes constant values.

tests/local_client_server_tcp_e2e/local_client_server_tcp_e2e.c
ewertons: Sure.

tests/local_client_server_tcp_e2e/local_client_server_tcp_e2e.c
dcristoloveanu: Then removing the initializers and doing a plain assignment is better, since otherwise there's 2 values set for the port.

---
In reply to: [109773444](https://github.com/Azure/azure-uamqp-c/pull/74#discussion_r109773444) [](ancestors = 109773444)

tests/local_client_server_tcp_e2e/local_client_server_tcp_e2e.c
dcristoloveanu: Actually I believe this statement is not true. The unit is only compiled once looking at Jenkins logs.

---
In reply to: [109773214](https://github.com/Azure/azure-uamqp-c/pull/74#discussion_r109773214) [](ancestors = 109773214)

ewertons:
Thanks for the review.

PR #76, opened 2017-04-08
Add float and double support for AMQP value encoding and decoding

dcristoloveanu:
Add float and double support for AMQP value encoding and decoding

### DIFF
--- a/src/amqpvalue.c
+++ b/src/amqpvalue.c
@@ -2674,6 +2674,60 @@ static int encode_long(AMQPVALUE_ENCODER_OUTPUT encoder_output, void* context, i
 	return result;
 }
 
+static int encode_float(AMQPVALUE_ENCODER_OUTPUT encoder_output, void* context, float value)
+{
+    int result;
+
+    uint32_t value_as_uint32 = *((uint32_t*)(void*)&value);
+    /* Codes_SRS_AMQPVALUE_01_289: [\<encoding name="ieee-754" code="0x72" category="fixed" width="4" label="IEEE 754-2008 binary32"/>] */
+    if ((output_byte(encoder_output, context, 0x72) != 0) ||
+        (output_byte(encoder_output, context, (value_as_uint32 >> 24) & 0xFF) != 0) ||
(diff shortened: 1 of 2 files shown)


src/amqpvalue.c
```diff
@@ -2674,6 +2674,60 @@ static int encode_long(AMQPVALUE_ENCODER_OUTPUT encoder_output, void* context, i
 	return result;
 }
 
+static int encode_float(AMQPVALUE_ENCODER_OUTPUT encoder_output, void* context, float value)
+{
+    int result;
+
+    uint32_t value_as_uint32 = *((uint32_t*)(void*)&value);
```
ewertons: This encoding is not following the IEEE 754 algorithm.
What about the fraction part?

src/amqpvalue.c
```diff
@@ -2674,6 +2674,60 @@ static int encode_long(AMQPVALUE_ENCODER_OUTPUT encoder_output, void* context, i
...
+    }
+
+    return result;
+}
+
+static int encode_double(AMQPVALUE_ENCODER_OUTPUT encoder_output, void* context, double value)
+{
+    int result;
+
+    uint64_t value_as_uint64 = *((uint64_t*)(void*)&value);
```
ewertons: Same here, how is the fraction part preserved?

src/amqpvalue.c
dcristoloveanu: As discussed, this simply takes the representation of the float and sends it.
Based on the C99 saying:
"Types
1 The C floating types match the IEC 60559 formats as follows:
— The float type matches the IEC 60559 single format.
— The double type matches the IEC 60559 double format."

Wikipedia says about IEC 60559 (https://en.wikipedia.org/wiki/IEEE_floating_point):
"The international standard ISO/IEC/IEEE 60559:2011 (with content identical to IEEE 754-2008) has been approved for adoption through JTC1/SC 25 under the ISO/IEEE PSDO Agreement[1] and published.[2]"

src/amqpvalue.c
dcristoloveanu: Same as above

PR #77, opened 2017-04-09
Add LogError to amqpvalue

dcristoloveanu:
Add LogError to amqpvalue. While adding LogError some cases where the calling code did not check whether it should call the amqpvalue code (for example the definitions.c file did not check for the number of items in a composite type, but relied on the fact that the amqpvalue_get_list_item would fail.

These cases should also be corrected now.
This required adding a amqpvalue_get_composite_item_count function in amqpvalue, which did not yet receive unit tests, but it will soon.

### DIFF
--- a/uamqp_generator/amqp_definitions_c.tt
+++ b/uamqp_generator/amqp_definitions_c.tt
@@ -181,8 +181,15 @@ int amqpvalue_get_<#= type_name #>(AMQP_VALUE value, <#= type_name.ToUpper() #>_
 		}
 		else
 		{
-			do
-			{
+            uint32_t list_item_count;
+            if (amqpvalue_get_list_item_count(list_value, &list_item_count) != 0)
+            {
+                result = __FAILURE__;
+            }
+            else
+            {
+			    do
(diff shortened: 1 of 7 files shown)


uamqp_generator/amqp_definitions_c.tt
```diff
@@ -194,71 +201,86 @@ int amqpvalue_get_<#= type_name #>(AMQP_VALUE value, <#= type_name.ToUpper() #>_
...
+				        if (item_value == NULL)
+				        {
 <# 					if (field.mandatory == "true") #>
 <# 					{ #>
-					{
-						<#= type_name #>_destroy(*<#= type_name.ToLower() #>_handle);
-						result = __FAILURE__;
-						break;
-					}
+					        {
```
ewertons: Are these spaces or tabs? should follow the style from the rest of the document.

dcristoloveanu: I'll probably replace tabs with spaces in one big swoop in another PR ...

PR #88, opened 2017-04-13
Replace strcpy with memcpy

dcristoloveanu:
Replace strcpy with memcpy

### DIFF
--- a/src/connection.c
+++ b/src/connection.c
@@ -957,7 +957,8 @@ CONNECTION_HANDLE connection_create2(XIO_HANDLE xio, const char* hostname, const
                 {
                     if (hostname != NULL)
                     {
-                        result->host_name = (char*)malloc(strlen(hostname) + 1);
+                        size_t hostname_length = strlen(hostname);
+                        result->host_name = (char*)malloc(hostname_length + 1);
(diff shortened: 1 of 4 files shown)


src/connection.c
```diff
@@ -978,7 +979,8 @@ CONNECTION_HANDLE connection_create2(XIO_HANDLE xio, const char* hostname, const
 
                     if (result != NULL)
                     {
-                        result->container_id = (char*)malloc(strlen(container_id) + 1);
+                        size_t container_id_length = strlen(container_id);
+                        result->container_id = (char*)malloc(container_id_length + 1);
```
ewertons: Would it be too conservative to use (sizeof(char) * container_id_length) ?

dcristoloveanu: Well, it needs to look for the end of the string and there is no need for it in this case, the length is known.

dcristoloveanu: sizeof(char) is always 1. C99 is very crisp about it.
"When applied to an operand that has type char, unsigned char, or signed char,
(or a qualified version thereof) the result is 1."

dcristoloveanu:
Cheers!

PR #89, opened 2017-04-14
Fixing on_delivery_settled to result in correct MESSAGE_SEND_RESULT value

ewertons:
Sorry, since unit tests are not present for message_sender and link it would be a big item to add for the fix I need.
I started the test suite for link.h though.

### DIFF
--- a/inc/azure_uamqp_c/link.h
+++ b/inc/azure_uamqp_c/link.h
@@ -19,20 +19,27 @@ typedef struct LINK_INSTANCE_TAG* LINK_HANDLE;
 
 typedef enum LINK_STATE_TAG
 {
-	LINK_STATE_DETACHED,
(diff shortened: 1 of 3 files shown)


tests/CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ add_subdirectory(amqp_management_ut)
 add_subdirectory(cbs_ut)
 add_subdirectory(connection_ut)
 add_subdirectory(frame_codec_ut)
+add_subdirectory(link_ut)
```
dcristoloveanu: Let's not adding here for UTs. Reason: this has to be rewritten with specs et all. The current added UTs do not have any requirements and give a false sense of safety.

ewertons:
Alright, thanks Dan.
I will remove the link_ut.

PR #90, opened 2017-04-15
Fix socket listener to listen on any IP address and also add LogErrors

dcristoloveanu:
Fix socket listener to listen on any IP address and also add LogErrors

### DIFF
--- a/src/socket_listener_win32.c
+++ b/src/socket_listener_win32.c
@@ -9,6 +9,7 @@
 #include "windows.h"
 #include "azure_c_shared_utility/optimize_size.h"
 #include "azure_c_shared_utility/socketio.h"
+#include "azure_c_shared_utility/xlogging.h"
 
 typedef struct SOCKET_LISTENER_INSTANCE_TAG
 {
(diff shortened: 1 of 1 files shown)


src/socket_listener_win32.c
```diff
@@ -138,31 +137,28 @@ int socketlistener_stop(SOCKET_LISTENER_HANDLE socket_listener)
...
-				{
-					socket_listener_instance->on_socket_accepted(socket_listener_instance->callback_context, io);
-				}
+                LogError("Could not create IO");
+                (void)closesocket(accepted_socket);
 			}
 			else
 			{
-				(void)closesocket(accepted_socket);
+				socket_listener->on_socket_accepted(socket_listener->callback_context, io);
```
ewertons: Since this is within socketio_create, shouldn't the cb be "on_socket_created" ?

dcristoloveanu: The event is on the fact that a socket was accepted. When a socket is accepted by the listener a new socketio XIO instance is created and handed to the user. So I really believe the name is good as is.

PR #93, opened 2017-04-15
Update SASL anonymous API to only expose the get interface function a…

dcristoloveanu:
Update SASL anonymous API to only expose the get interface function and minor specs and tests updates for SASL anonymous

### DIFF
--- a/src/sasl_anonymous.c
+++ b/src/sasl_anonymous.c
@@ -13,49 +13,43 @@ typedef struct SASL_ANONYMOUS_INSTANCE_TAG
 	unsigned char dummy;
 } SASL_ANONYMOUS_INSTANCE;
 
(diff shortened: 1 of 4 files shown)


src/sasl_anonymous.c
```diff
@@ -112,17 +106,27 @@ int saslanonymous_challenge(CONCRETE_SASL_MECHANISM_HANDLE concrete_sasl_mechani
 	}
 	else
 	{
-		/* Codes_SRS_SASL_ANONYMOUS_01_013: [saslanonymous_challenge shall set the response_bytes buffer to NULL and 0 size as the ANONYMOUS SASL mechanism does not implement challenge/response.] */
+		/* Codes_SRS_SASL_ANONYMOUS_01_013: [`saslanonymous_challenge` shall set the `buffer` field to NULL and `size` to 0 in the `response_bytes` argument as the ANONYMOUS SASL mechanism does not implement challenge/response.] */
 		response_bytes->bytes = NULL;
```
ewertons: These lines are showing as removed. Are they really missing?

dcristoloveanu: We're talking about replacement of Codes_01_013 with Codes_01_013 the new one. I do not understand what you thing is removed ...

---
In reply to: [111764122](https://github.com/Azure/azure-uamqp-c/pull/93#discussion_r111764122) [](ancestors = 111764122)

PR #94, opened 2017-04-15
Update dependnecies so that --recursive should not be needed anymore

dcristoloveanu:
Update dependnecies so that --recursive should not be needed anymore.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,12 @@
-[submodule "azure-c-shared-utility"]
-	path = c-utility
-	url = https://github.com/Azure/azure-c-shared-utility.git
+[submodule "deps/azure-c-shared-utility"]
+	path = deps/azure-c-shared-utility
+	url = https://www.github.com/Azure/azure-c-shared-utility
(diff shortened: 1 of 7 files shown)


dcristoloveanu:
I beg to disagree. Each module should simply pull in and build a flat list of dependencies. In that way you solve several problems:

1. It is controllable when the submodule is added to the CMakeLists
2. No need for recursive anymore.

This change was  desired since a long time by me, Andrei, Marc Schier, etc.
It should actually be applied to iot-sdk-c also (and GW).

The duplication you talk about will not happen anymore since the recursive update is not needed anymore. If you do insist on --recursive update then you are simply choosing by yourself to do the duplication, aka not my problem anymore :-).

ewertons:
So the plan is to apply the same to all the modules, including azure-iot-sdk-c?

If that is the case, I didn't get to know that when you planned to make this change.
And being so, please go ahead with the merge.

I will need to sync with Tamer since this will probably impact the release scripts as well.

Sent from my Windows Phone
________________________________

PR #99, opened 2017-04-20
Change the status of link and message_sender before bubbling up pending deliveries

ewertons:


dcristoloveanu:
>             set_message_sender_state(message_sender_instance, MESSAGE_SENDER_STATE_IDLE);

I think this should get revered here too 

---
Refers to: src/message_sender.c:562 in 7fa6295. [](commit_id = 7fa629554a60d89191762c8efc352c6d7aa59045, deletion_comment = False)

dcristoloveanu:
:shipit:

ewertons:
Yes, makes sense. I didn't change it thinking it wouldn't be necessary, but it better follow the same order as the other statements.
I'll make the change.

PR #105, opened 2017-05-02
Beautify specs for sasl_plain and nitpick update on tests

dcristoloveanu:
Beautify specs for sasl_plain and nitpick update on tests


tameraw:
:shipit:

PR #107, opened 2017-05-04
Fix memory leaks

zolvarga:
Fixed memory leaks for failure cases in AMQP definitions

Best Regards,
Zoltan


dcristoloveanu:
:shipit:

PR #116, opened 2017-05-22
Improve performance by using ref counting on amqpvalue.

dcristoloveanu:
Improve performance by using ref counting on amqpvalue.


et q
```diff
@@ -0,0 +1,4354 @@
+[33mcommit 4485acd3f042fba09dbafad21bdf3c79dd6dd452[m
```
ewertons: This shouldn't be checked in, right?

PR #117, opened 2017-05-22
Add extra Ubuntu 16.04 build

dcristoloveanu:
Add extra Ubuntu 16.04 build


jenkins/ubuntu1604_c.sh
```diff
@@ -0,0 +1,28 @@
...
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+#
+
+set -e
+
+script_dir=$(cd "$(dirname "$0")" && pwd)
+build_root=$(cd "${script_dir}/.." && pwd)
+log_dir=$build_root
+make_install=
+build_folder=$build_root"/cmake"
```
BertKleewein: Don't we want all our builds to happen in subfolders underneath cmake?

PR #118, opened 2017-05-23
Add ref counting to message also

dcristoloveanu:
Add ref counting to message also.
Another PR is pending for message UTs which should add coverage for that module.

### DIFF
--- a/src/message.c
+++ b/src/message.c
@@ -7,6 +7,7 @@
 #include "azure_c_shared_utility/optimize_size.h"
 #include "azure_c_shared_utility/gballoc.h"
 #include "azure_c_shared_utility/xlogging.h"
+#include "azure_c_shared_utility/refcount.h"
 #include "azure_uamqp_c/message.h"
 #include "azure_uamqp_c/amqpvalue.h"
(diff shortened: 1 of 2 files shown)


src/message.c
```diff
@@ -107,212 +110,52 @@ MESSAGE_HANDLE message_clone(MESSAGE_HANDLE source_message)
...
-			if ((result != NULL) && (source_message->properties != NULL))
+void message_destroy(MESSAGE_HANDLE message)
+{
+	if (message == NULL)
+	{
+		LogError("NULL message");
+	}
+	else
+	{
+		if (DEC_REF(MESSAGE_INSTANCE, message) == DEC_RETURN_ZERO)
```
jebrando: >if (DEC_REF(MESSAGE_INSTANCE, message)  [](start = 2, length = 39)

More Thought required.

jebrando:
:clock1:

dcristoloveanu:
Thank you for the review, I will leave this out for the moment. I need to think through a bit more how I want to position ref counting vs cloning of messages.

PR #119, opened 2017-05-23
Make all expected calls be STRICT_

dcristoloveanu:
Make all expected calls be STRICT_

### DIFF
--- a/tests/amqpvalue_ut/amqpvalue_ut.c
+++ b/tests/amqpvalue_ut/amqpvalue_ut.c
@@ -210,7 +210,7 @@ TEST_FUNCTION(amqpvalue_create_null_succeeds)
 {
     // arrange
 	AMQP_VALUE result;
-    EXPECTED_CALL(gballoc_malloc(IGNORED_NUM_ARG));
+    STRICT_EXPECTED_CALL(gballoc_malloc(IGNORED_NUM_ARG));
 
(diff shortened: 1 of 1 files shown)


tests/amqpvalue_ut/amqpvalue_ut.c
```diff
@@ -210,7 +210,7 @@ TEST_FUNCTION(amqpvalue_create_null_succeeds)
 {
     // arrange
 	AMQP_VALUE result;
-    EXPECTED_CALL(gballoc_malloc(IGNORED_NUM_ARG));
+    STRICT_EXPECTED_CALL(gballoc_malloc(IGNORED_NUM_ARG));
```
ewertons: Looks like most of the changes were for calls that were already ignoring the arguments.
Not sure what is the point of the changes, are you planning on deprecating EXPECTED_CALL(...) ?

dcristoloveanu: At some point I decided I should only use STRICT_. Because otherwise it is much easier to err on the side of not validating a parameter. And the fact that IGNORE_NUM_ARG and IGNORE_PTR_ARG do not force writing an extra line makes it the same to use STRICT_EXPECTED_CALL or EXPECTED_CALL from the point of view of amount of code.

PR #130, opened 2017-06-12
Patch multiple value handling to allow single SASL mechanisms in SASL_init

### LINKED ISSUE #129: When SASL Mechanisms is sent with 1 SASL mechanism client fails
saslclientio treats the sasl-server-mechanisms in the sasl_mechanisms performative as an array (mandatory).
That is not correct.

The field sasl-server-mechanisms is multiple symbol, so it can be simply one symbol.

dcristoloveanu:
The fix creates an array whenever dealing with multiple values in the getters for the fields.
An E2E test will be following soon testing with one SASL mechanism. Unfortunately such a test requires a saslserverio (a hacked up flavor will be created in order to be able to do the tests).

This should fix #129.


jebrando:
> // Copyright (c) Microsoft. All rights reserved.

amqpvalue_create_array needs to initialize the array_value.count and array_value.items.

---
Refers to: src/amqp_definitions.c:3 in 8cd30be. [](commit_id = 8cd30be40b82a6b58580f49f5c461d75def889c0, deletion_comment = False)

jebrando:
> #include <stdbool.h>

Are you going to put the pragma warning back in?

---
Refers to: uamqp_generator/amqp_definitions_c.tt:19 in 8cd30be. [](commit_id = 8cd30be40b82a6b58580f49f5c461d75def889c0, deletion_comment = False)

jebrando:
:shipit:

dcristoloveanu:
> // Copyright (c) Microsoft. All rights reserved.

Thanks, fixed!

---
In reply to: [307877599](https://github.com/Azure/azure-uamqp-c/pull/130#issuecomment-307877599) [](ancestors = 307877599)

---
Refers to: src/amqp_definitions.c:3 in 8cd30be. [](commit_id = 8cd30be40b82a6b58580f49f5c461d75def889c0, deletion_comment = False)

dcristoloveanu:
> #include <stdbool.h>

Actually it is not needed once the while (0) is fixed to be while while((void)0,0)

---
In reply to: [307878113](https://github.com/Azure/azure-uamqp-c/pull/130#issuecomment-307878113) [](ancestors = 307878113)

---
Refers to: uamqp_generator/amqp_definitions_c.tt:19 in 8cd30be. [](commit_id = 8cd30be40b82a6b58580f49f5c461d75def889c0, deletion_comment = False)

PR #135, opened 2017-06-12
Bulk convert tasb to spaces

dcristoloveanu:
Bulk convert tabs to spaces


jebrando:
> // Copyright (c) Microsoft. All rights reserved.

This file doesn't look like it changed anything, but since CodeFlow sometimes doesn't pick up things correctly, this comment may not be accurate.

---
Refers to: inc/azure_uamqp_c/amqpvalue.h:1 in 940033f. [](commit_id = 940033f6a33bf29583bd339a80e51f86c09b8231, deletion_comment = False)

src/frame_codec.c
```diff
@@ -14,132 +14,132 @@
 #include "azure_uamqp_c/amqpvalue.h"
 
 #define FRAME_HEADER_SIZE 8
-#define MAX_TYPE_SPECIFIC_SIZE	((255 * 4) - 6)
```
jebrando: Do we need something in Jenkins to do this automagically?

jebrando:
:shipit:

src/frame_codec.c
dcristoloveanu: Yeah, we could have a script that does this auto-magically. I would love that. But did not put much thought into that yet.

dcristoloveanu:
> // Copyright (c) Microsoft. All rights reserved.

Yeah, I did a bulk replace, so did not check each individual file....

---
In reply to: [307855898](https://github.com/Azure/azure-uamqp-c/pull/135#issuecomment-307855898) [](ancestors = 307855898)

---
Refers to: inc/azure_uamqp_c/amqpvalue.h:1 in 940033f. [](commit_id = 940033f6a33bf29583bd339a80e51f86c09b8231, deletion_comment = False)

Issue #139, opened 2017-06-13
amqpvalue.c reports unnecessary error


jebrando:
amqpvalue_get_sasl_mechanisms (Line 10167) attempts to get the array or symbol and will report an error even though the operation succeeds when the type is a symbol.

dcristoloveanu:
@jebrando I'll take a look at a fix.

dcristoloveanu:
This has been fixed and integrated in master.

PR #140, opened 2017-06-15
Fix unecessary error when type is not ARRAY

dcristoloveanu:
This should fix the unnecessary warning when decoding single mechanism values in the SASL mechanisms array.


jebrando:
:shipit:

dcristoloveanu:
This will fix #139.

Issue #132, opened 2017-06-12
header detect IO needs rewriting


dcristoloveanu:
- With tests
- Multiple headers should be possible to be detected (SASL, AMQP). This is needed to implement SASL on server side.

dcristoloveanu:
This has been done.

PR #141, opened 2017-06-21
Rewrite header detect IO with tests to allow future development of SASL server IO

dcristoloveanu:
This rewrites header detect IO in order to be able to detect a sequence of stacked headers.
Essentially now it should be possible to have a detection of an 3.1.0.0 header followed by another detection of a 3.1.0.0 header and then followed by another header and so on.
The sequence always ends in a 0.1.0.0 header though.

Tests have also been written for the module.

The interaction with the detected IOs (SASL for example) is done by placing in the Create parameters a pointer to an ON_BYTES_RECEIVED callback and its context.


dcristoloveanu:
This covers issue #132.

PR #157, opened 2017-07-07
Do not delete messages in on_link_state_changed on detach or error

### LINKED ISSUE #156: Usage after free when link is detached or set to error state
* OS and version used: Windows 10
* Version: master

# Description of the issue

In the event of a session going to either discarding or error state messages stored will be used after free and then double-freed. This can be seen by running the iothub_client_sample_amqp and disconnect the computer from network while the messages are being transmitted, but before they have been completed. The result will be a seg-fault.

# Code analysis

The problem starts in `on_session_state_changed` in `link.c`. When `new_session_state` is either `SESSION_STATE_DISCARDING` or `SESSION_STATE_ERROR` `set_link_state` will be called with the new link state (either state detached or state error). This will cause the callback `on_link_state_changed` to be called for all stored link instances. The actual function that is called will  be `on_link_state_changed` in `message_sender.c`. In all cases a call to `indicate_all_messages_as_error` will be called which will destroy stored messages.

After `set_link_state` is finished `on_session_state_changed` will call `remove_all_pending_deliveries` which will call the `on_delivery_settled` callback on all pending deliveries. This callback actually calls `on_delivery_settled` in `message_sender.c` with context being a pointer to a MESSAGE_WITH_CALLBACK that was previously freed in `indicate_all_messages_as_error`. `on_delivery_settled` will attempt to use this pointer which results in a crash.

# Fix

As far as I can tell the best way to fix this is to remove the two offneding calls to `indicate_all_messages_as_error`. I can not see that this function does anything that will not be done by the call chain down to `on_delivery_settled`. However `remove_all_pending_deliveries` will free additional resources as well as the messages so this must at least be called.

Shakti213:
Messages will be deleted when pending deliveries are deleted instead.

This should fix #156

### DIFF
--- a/src/message_sender.c
+++ b/src/message_sender.c
@@ -562,6 +562,7 @@ static void on_link_state_changed(void* context, LINK_STATE new_link_state, LINK
         {
             /* User initiated transition, we should be good */
             set_message_sender_state(message_sender_instance, MESSAGE_SENDER_STATE_IDLE);
+            indicate_all_messages_as_error(message_sender_instance);
         }
         else if (message_sender_instance->message_sender_state != MESSAGE_SENDER_STATE_IDLE)
         {
@@ -573,6 +574,7 @@ static void on_link_state_changed(void* context, LINK_STATE new_link_state, LINK
         if (message_sender_instance->message_sender_state != MESSAGE_SENDER_STATE_ERROR)
         {
             set_message_sender_state(message_sender_instance, MESSAGE_SENDER_STATE_ERROR);
+            indicate_all_messages_as_error(message_sender_instance);
         }
         break;
     }
--- a/src/message_sender.c
+++ b/src/message_sender.c
@@ -562,7 +562,6 @@ static void on_link_state_changed(void* context, LINK_STATE new_link_state, LINK
         {
             /* User initiated transition, we should be good */
             set_message_sender_state(message_sender_instance, MESSAGE_SENDER_STATE_IDLE);
-            indicate_all_messages_as_error(message_sender_instance);
         }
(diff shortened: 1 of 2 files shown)


src/message_sender.c
```diff
@@ -562,7 +562,6 @@ static void on_link_state_changed(void* context, LINK_STATE new_link_state, LINK
         {
             /* User initiated transition, we should be good */
             set_message_sender_state(message_sender_instance, MESSAGE_SENDER_STATE_IDLE);
-            indicate_all_messages_as_error(message_sender_instance);
```
dcristoloveanu: >indicate_all_messages_as_error [](start = 12, length = 30)

Thanks for looking into this.
I looked at the change and at what we have in link.c and I think we should go one step further to cover as much as we can of the cases.

I think 
remove_all_pending_deliveries(link_instance, true);
should be called whenever the link state is set to DETACHED/ERROR.

That would be link.c ln 280, ln 330, ln. 549, ln. 555.

Or directly in the set_link_state if the link_state is DETACHED or ERROR.

In that case the indicate_all_messages_as_error call actually can stay, as it won't hurt, but probably will be hit with an empty list of messages.

I would have made the change by myself, but I wanted to ask you if you can check that it also solves your case.

dcristoloveanu: @Shakti213 Let me know if you want to push a new commit and I'll have a look and merge it in. Or otherwise I can create a fix that you can have a look at.

---
In reply to: [126278896](https://github.com/Azure/azure-uamqp-c/pull/157#discussion_r126278896) [](ancestors = 126278896)

Shakti213: Hi, I can check if this solves my issue as well and update with a new commit afterwards.

dcristoloveanu:
:shipit:

Issue #148, opened 2017-06-30
Remove usage of cstdbool in headers


dcristoloveanu:
Because bool is a built in type in C++, it is not necessary to include cstdbool.

dcristoloveanu:
This has been addressed.

PR #159, opened 2017-07-14
Remove #include of <cstdbool>

dcristoloveanu:
Remove #include of <cstdbool>


dcristoloveanu:
This addresses #148.

tameraw:
:shipit:

PR #160, opened 2017-07-14
Remove const qualifier for AMQP_VALUE

### LINKED ISSUE #155: Usage of const AMQP_VALUE should be removed
There are several places where const AMQP_VALUE is used. See amqp_frame codec for example.
These usages should be changed to AMQP_VALUE, as there is no meaning to applying const in this case.

dcristoloveanu:
Fixes #155. Remove const qualifier for AMQP_VALUE.


tameraw:
:shipit:

PR #173, opened 2017-08-01
Fix memory leak on send_one_message()

ewertons:
Fix a few spots leading to memory leaks in message_sender.


dcristoloveanu:
:shipit:

ewertons:
Awesome, thanks!

PR #185, opened 2017-10-02
Add async operation and cancel for message_sender_send_async

dcristoloveanu:
Add async operation and cancel for message_sender_send_async


src/async_operation.c
```diff
@@ -0,0 +1,87 @@
...
+        if (async_operation == NULL)
+        {
+            /* Codes_SRS_ASYNC_OPERATION_01_004: [ If allocating memory for the new asynchronous operation instance fails, `async_operation_create` shall fail and return NULL.]*/
+            LogError("Cannot allocate memory for async operation");
+        }
+        else
+        {
+            /* Codes_SRS_ASYNC_OPERATION_01_001: [ `async_operation_create` shall return a non-NULL handle to a newly created asynchronous operation instance.]*/
+            async_operation->async_operation_cancel_handler = async_operation_cancel_handler;
+            goto all_ok;
```
jebrando: >goto all_ok; [](start = 12, length = 12)

I don't think this goto is providing much value.  without it you just have to have async_operation = NULL in two more places.  There's not any deallocation or save by it.

jebrando:
>         LOG(AZ_LOG_TRACE, 0, "%s", (value_as_string = amqpvalue_to_string(value)));

If value_as_string is NULL then the log will be interesting.   I know on windows it will print the word null, but on other platforms?  Should you move the LOG inside the if?

---
Refers to: src/message_sender.c:170 in 60489b4. [](commit_id = 60489b45d67a8b866bcca49e0563c96b78fd1d7f, deletion_comment = False)

dcristoloveanu:
>         LOG(AZ_LOG_TRACE, 0, "%s", (value_as_string = amqpvalue_to_string(value)));

Fixed by using P_OR_NULL.

---
In reply to: [333647033](https://github.com/Azure/azure-uamqp-c/pull/185#issuecomment-333647033) [](ancestors = 333647033)

---
Refers to: src/message_sender.c:170 in 60489b4. [](commit_id = 60489b45d67a8b866bcca49e0563c96b78fd1d7f, deletion_comment = False)

src/async_operation.c
dcristoloveanu: It can go either way. I removed it anyhow this time :-).

---
In reply to: [142210232](https://github.com/Azure/azure-uamqp-c/pull/185#discussion_r142210232) [](ancestors = 142210232)

Issue #188, opened 2017-10-20
double free on message_sender if messagesender_send_async fails with SEND_ONE_MESSAGE_ERROR


ewertons:
This was introduced by the recent changes regarding async operations in uAMQP.
It broke the reconnection logic in the Azure IoT C SDK.

**Using commit:** [master/10-20-2017](https://github.com/Azure/azure-uamqp-c/commit/5b3db494254b5b17ac4e7e4ceefc43150f7564b1)

**Repro steps:** same as in https://github.com/Azure/azure-iot-sdk-c/issues/251

**Root cause:**
The current implementation calls [remove_pending_message_by_index](https://github.com/Azure/azure-uamqp-c/blob/5b3db494254b5b17ac4e7e4ceefc43150f7564b1/src/message_sender.c#L899) followed by [async_operation_destroy](https://github.com/Azure/azure-uamqp-c/blob/5b3db494254b5b17ac4e7e4ceefc43150f7564b1/src/message_sender.c#L900) if send_one_message fails (within messagesender_send_async).
However remove_pending_message_by_index already calls [async_operation_destroy](https://github.com/Azure/azure-uamqp-c/blob/5b3db494254b5b17ac4e7e4ceefc43150f7564b1/src/message_sender.c#L62) for the respective ASYNC_OPERATION_HANDLE, resulting in a double free.

**Traces/Logs:**
```
-> [TRANSFER]* {0,232,<E9 00 00 00>,2147563264,false,false}
IoTHubClient_LL_SendEventAsync accepted message [243] for transmission to IoT Hub.
-> [TRANSFER]* {0,233,<EA 00 00 00>,2147563264,false,false}
IoTHubClient_LL_SendEventAsync accepted message [244] for transmission to IoT Hub.
-> [TRANSFER]* {0,234,<EB 00 00 00>,2147563264,false,false}
IoTHubClient_LL_SendEventAsync accepted message [245]
Breakpoint 1, async_operation_destroy (async_operation=0x177888) at /home/pi/azure-iot-sdk-c/uamqp/src/async_operation.c:54
54	    if (async_operation == NULL)
async_operation = 0x177888
__func__ = "async_operation_destroy"
#0  async_operation_destroy (async_operation=0x177888) at /home/pi/azure-iot-sdk-c/uamqp/src/async_operation.c:54
[...]

PR #189, opened 2017-10-21
Fix double-free crash if send_one_message fails with SEND_ONE_MESSAGE_ERROR (github issue #188)

ewertons:
Addresses https://github.com/Azure/azure-uamqp-c/issues/188

Fix has been verified with real reconnection scenario on a Linux device.

This module doesn't have unit tests (correct me if I'm wrong), so no other changes were added to this PR.


dcristoloveanu:
Thanks for fixing this.

PR #192, opened 2017-10-22
Make link_transfer return an ASYNC_OPERATION_HANDLE

dcristoloveanu:
Make link_transfer return an ASYNC_OPERATION_HANDLE so that subsequently it can be plumbed into the message_sender_send_async to have a clean timeouts

### DIFF
--- a/src/link.c
+++ b/src/link.c
@@ -15,6 +15,7 @@
 #include "azure_uamqp_c/amqpvalue.h"
 #include "azure_uamqp_c/amqp_definitions.h"
 #include "azure_uamqp_c/amqp_frame_codec.h"
+#include "azure_uamqp_c/async_operation.h"
 
 #define DEFAULT_LINK_CREDIT 10000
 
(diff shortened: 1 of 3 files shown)


src/link.c
```diff
@@ -607,14 +614,15 @@ static void on_session_flow_on(void* context)
 static void on_send_complete(void* context, IO_SEND_RESULT send_result)
 {
     LIST_ITEM_HANDLE delivery_instance_list_item = (LIST_ITEM_HANDLE)context;
-    DELIVERY_INSTANCE* delivery_instance = (DELIVERY_INSTANCE*)singlylinkedlist_item_get_value(delivery_instance_list_item);
-    LINK_INSTANCE* link_instance = (LINK_INSTANCE*)delivery_instance->link;
+    ASYNC_OPERATION_HANDLE pending_delivery_operation = (ASYNC_OPERATION_HANDLE)singlylinkedlist_item_get_value(delivery_instance_list_item);
+    DELIVERY_INSTANCE* delivery_instance = (DELIVERY_INSTANCE*)GET_ASYNC_OPERATION_CONTEXT(DELIVERY_INSTANCE, pending_delivery_operation);
```
jebrando: >DELIVERY_INSTANCE* delivery_instance = (DELIVERY_INSTANCE*) [](start = 4, length = 59)

I'm unfamiliar with the API, so forgive me for my ignorance.  Should you check for NULL here?

dcristoloveanu: Not needed as it is put in by the same module.

Issue #197, opened 2017-11-28
SEGFAULT on link_transfer_async when session_send_transfer fails with SESSION_SEND_TRANSFER_ERROR


ewertons:
This issue was experienced using azure-iot-sdk-c (1.1.28).

Environment: Raspbian on Pi3 (Linux raspberrypi 4.9.35-v7+ #1014 SMP Fri Jun 30 14:47:43 BST 2017 armv7l GNU/Linux)

Here are the repro steps:
1.	Run sample https://github.com/Azure/azure-iot-sdk-c/tree/master/iothub_client/samples/iothub_client_sample_x509
- Sample was modified to send 5000+ messages, one per second (attached).
2.	Let the first messages succeed to be sent;
3.	Unplug network cable
4.	Wait for 4 minutes
5.	Reconnection is triggered, and crash occurs.

Callstack:

#0  __GI___libc_free (mem=0x16dd7c) at malloc.c:2945
#1  0x000bba8c in  (link=0x132150, message_format=2147563264, payloads=0x7efeb524, payload_count=1, on_delivery_settled=0xc0e44 <on_delivery_settled>, callback_context=0x191d20, link_transfer_error=0x7efeb510, timeout=0)
    at /home/ewertons/azure-iot-sdk-c/uamqp/src/link.c:1252
#2  0x000c2250 in send_one_message (message_sender=0x12cd78, pending_send=0x191d20, message=0x191c38) at /home/ewertons/azure-iot-sdk-c/uamqp/src/message_sender.c:521
#3  0x000c31e4 in messagesender_send_async (message_sender=0x12cd78, message=0x191c38, on_message_send_complete=0x70628 <internal_on_event_send_complete_callback>, callback_context=0x191a28, timeout=0)
    at /home/ewertons/azure-iot-sdk-c/uamqp/src/message_sender.c:898
#4  0x00070aac in send_batched_message_and_reset_state (instance=0x12d2b8, send_pending_events_state=0x7efeb6a0) at /home/ewertons/azure-iot-sdk-c/iothub_client/src/iothubtransport_amqp_telemetry_messenger.c:1133
#5  0x00071228 in send_pending_events (instance=0x12d2b8) at /home/ewertons/azure-iot-sdk-c/iothub_client/src/iothubtransport_amqp_telemetry_messenger.c:1279
[...]

PR #202, opened 2017-11-28
Fix for segfault on link.c due to free on async operation content (#197)

ewertons:
This is to address issue #197

### DIFF
--- a/src/link.c
+++ b/src/link.c
@@ -1124,9 +1124,10 @@ static void link_transfer_cancel_handler(ASYNC_OPERATION_HANDLE link_transfer_op
     {
         pending_delivery->on_delivery_settled(pending_delivery->callback_context, pending_delivery->delivery_id, LINK_DELIVERY_SETTLE_REASON_CANCELLED, NULL);
     }
-    free(pending_delivery);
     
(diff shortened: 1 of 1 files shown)


src/link.c
```diff
@@ -1124,9 +1124,10 @@ static void link_transfer_cancel_handler(ASYNC_OPERATION_HANDLE link_transfer_op
     {
         pending_delivery->on_delivery_settled(pending_delivery->callback_context, pending_delivery->delivery_id, LINK_DELIVERY_SETTLE_REASON_CANCELLED, NULL);
     }
-    free(pending_delivery);
     
     (void)singlylinkedlist_remove_if(((LINK_HANDLE)pending_delivery->link)->pending_deliveries, remove_pending_delivery_condition_function, pending_delivery);
+
+    async_operation_destroy(link_transfer_operation);
```
ewertons: @dcristoloveanu , based on our conversation I understand the async_operation_destroy shall be invoked from within the cancellation handler.
Is that understanding correct?

I see uamqp doesn't call async_operation_cancel anywhere yet. We need to make sure this is well verified when it starts to use it.

dcristoloveanu: Your understanding is correct, thanks for fixing this.

---
In reply to: [153641131](https://github.com/Azure/azure-uamqp-c/pull/202#discussion_r153641131) [](ancestors = 153641131)

PR #203, opened 2018-01-17
Always provide on_send_complete callback to xio_send

JetstreamRoySprowl:


src/connection.c
```diff
@@ -102,13 +102,21 @@ static void connection_set_state(CONNECTION_HANDLE connection, CONNECTION_STATE
     }
 }
 
+// This callback usage needs to be either verified and commented or integrated into
```
ewertons: Maybe you should write // TODO: ...
That helps track what is pending.
Might need to get Dan's opinion on this as well.

ewertons: IF you end up adding the "TODO:" then add it everywhere you made such comment.

dcristoloveanu: Could you file an issue in GitHub also so it gets tracked? Thanks!

---
In reply to: [162208823](https://github.com/Azure/azure-uamqp-c/pull/203#discussion_r162208823) [](ancestors = 162208823)

dcristoloveanu: One issue for all the occurrences should be enough

---
In reply to: [162209233](https://github.com/Azure/azure-uamqp-c/pull/203#discussion_r162209233) [](ancestors = 162209233,162208823)

samples/local_server_tls_sample/tls_server_io_schannel.c
```diff
@@ -464,6 +464,14 @@ static int internal_send(TLS_IO_INSTANCE* tls_io_instance, const void* buffer, s
     return result;
 }
 
+
+// A simple sample callback.
+static void sample_on_send_complete(void* context, IO_SEND_RESULT send_result)
```
dcristoloveanu: >sample_ [](start = 12, length = 7)

Can you please remove the sample prefix as this might move to src at some point. I'd rather have it unckecked_

JetstreamRoySprowl: "I'd rather have it unckecked_"... Will do.

Issue #208, opened 2018-02-06
AMQP management does not bubble up the message handle for the response


dcristoloveanu:
Giving the message handle to the user is the easiest way to expose all what is needed for parsing the response information.

Thus the callback for execute operation complete should be:

typedef void(*ON_AMQP_MANAGEMENT_EXECUTE_OPERATION_COMPLETE)(void* context, AMQP_MANAGEMENT_EXECUTE_OPERATION_RESULT execute_operation_result, unsigned int status_code, const char* status_description, MESSAGE_HANDLE message);

dcristoloveanu:
Labeled as bug since it blocks building other AMQP management operations than CBS.

PR #209, opened 2018-02-07
Bubble up the message that was received as an AMQP management reply

dcristoloveanu:
Addresses issue #208


jspaith:
> # amqp_management requirements

Check c-shared version & its changes.

---
Refers to: devdoc/amqp_management_requirements.md:1 in 332abf7. [](commit_id = 332abf7b2ef95832e9589dcebbc925f78fae9531, deletion_comment = False)

Issue #210, opened 2018-02-08
AMQP management module should use statusCode and statusDescription


dcristoloveanu:
AMQP management module should use statusCode and statusDescription. Also in order to allow CBS to function, because CBS uses different key names, we need to provide setters to override these values.

PR #211, opened 2018-02-08
Update AMQP management module to use statusCode and statusDescription…

dcristoloveanu:
Update AMQP management module to use statusCode and statusDescription by default.
Also allow overriding these keynames with different values so that CBS can use them.
Update specs, code and tests for AMQP management and CBS modules.
CBS module now calls the setters to override the default values, so that the behavior remains unchanged at that level.
There was some minor restructuring done to the _create functions to make sure that there are no leaks.

This addresses #210.

### DIFF
--- a/src/amqp_management.c
+++ b/src/amqp_management.c
@@ -9,6 +9,7 @@
 #include "azure_c_shared_utility/gballoc.h"
 #include "azure_c_shared_utility/singlylinkedlist.h"
 #include "azure_c_shared_utility/xlogging.h"
+#include "azure_c_shared_utility/crt_abstractions.h"
 #include "azure_uamqp_c/amqp_management.h"
 #include "azure_uamqp_c/link.h"
 #include "azure_uamqp_c/message_sender.h"
@@ -48,6 +49,8 @@ typedef struct AMQP_MANAGEMENT_INSTANCE_TAG
     ON_AMQP_MANAGEMENT_ERROR on_amqp_management_error;
     void* on_amqp_management_error_context;
     AMQP_MANAGEMENT_STATE amqp_management_state;
+    char* status_code_key_name;
+    char* status_description_key_name;
     int sender_connected : 1;
     int receiver_connected : 1;
 } AMQP_MANAGEMENT_INSTANCE;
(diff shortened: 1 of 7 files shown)


src/cbs.c
```diff
@@ -256,54 +256,79 @@ static void on_amqp_management_execute_operation_complete(void* context, AMQP_MA
...
+                        }
+                    }
+
+                    amqp_management_destroy(cbs->amqp_management);
                 }
+
+                singlylinkedlist_destroy(cbs->pending_operations);
             }
+
+            free(cbs);
```
jspaith: >free(cbs); [](start = 12, length = 10)

Nit - instead of individually freeing up members above in error case, could you just invoke cbs_destroy() and let it hide all the details?

src/amqp_management.c
```diff
@@ -597,190 +600,252 @@ static int add_string_key_value_pair_to_map(AMQP_VALUE map, const char* key, con
...
-                                free(receiver_link_name);
+                                amqpvalue_destroy(target);
                             }
 
-                            free(sender_link_name);
+                            amqpvalue_destroy(source);
                         }
 
-                        amqpvalue_destroy(target);
+                        free(amqp_management->status_description_key_name);
```
jspaith: >free(amqp_management->status_description_key_name); [](start = 24, length = 51)

Nit - along lines of other - could a single call in error case to amqp_management_destroy() be used here on those frees that are tied to the amqp_management struct?

src/cbs.c
dcristoloveanu: I prefer writing it this way. The reason is that it is easier for me to track the frees in the same method. And if I were to use the call to destroy for cleanup of create then I'd have to make sure that every member gets initialized with NULL prior to executing the code.

Not saying it cannot be done, but the result is pretty much the same (and tests have to check both ways of writing it it anyhow), thus changing it would be 0 ROI.

---
In reply to: [167072243](https://github.com/Azure/azure-uamqp-c/pull/211#discussion_r167072243) [](ancestors = 167072243)

src/amqp_management.c
dcristoloveanu: Same answer like for the other comment.

---
In reply to: [167072806](https://github.com/Azure/azure-uamqp-c/pull/211#discussion_r167072806) [](ancestors = 167072806)

dcristoloveanu:
@annatisch @jspaith 
Thanks for the review!

PR #212, opened 2018-02-14
Split definitions h

jspaith:
The existing amqp_definitions.h describes the entire AMQP protocol as auto-generated from .tt files.  For the vast majority of consumers, they need a much, much smaller subset of these headers.

Split up the .tt such that each AMQP type gets its own header file which can be selectively included.

Also create <..>_internal.h headers for link/message/session.  Session/link/message.h are widely included even outside this repo.  The functions in _internal.h are (a) only used internally by uamqp itself but more importantly (b) bring in **TONS** of amqp definitions which would've been a big hit to all the callers.  Splitting lets consumers of session.h et al keep perf benefits.


dcristoloveanu:
Thanks John! I bet this was a fun change to make.

inc/azure_uamqp_c/session.h
```diff
@@ -43,24 +45,11 @@ DEFINE_ENUM(SESSION_SEND_TRANSFER_RESULT, SESSION_SEND_TRANSFER_RESULT_VALUES)
     typedef bool(*ON_LINK_ATTACHED)(void* context, LINK_ENDPOINT_HANDLE new_link_endpoint, const char* name, role role, AMQP_VALUE source, AMQP_VALUE target);
 
     MOCKABLE_FUNCTION(, SESSION_HANDLE, session_create, CONNECTION_HANDLE, connection, ON_LINK_ATTACHED, on_link_attached, void*, callback_context);
-    MOCKABLE_FUNCTION(, SESSION_HANDLE, session_create_from_endpoint, CONNECTION_HANDLE, connection, ENDPOINT_HANDLE, connection_endpoint, ON_LINK_ATTACHED, on_link_attached, void*, callback_context);
     MOCKABLE_FUNCTION(, int, session_set_incoming_window, SESSION_HANDLE, session, uint32_t, incoming_window);
     MOCKABLE_FUNCTION(, int, session_get_incoming_window, SESSION_HANDLE, session, uint32_t*, incoming_window);
     MOCKABLE_FUNCTION(, int, session_set_outgoing_window, SESSION_HANDLE, session, uint32_t, outgoing_window);
     MOCKABLE_FUNCTION(, int, session_get_outgoing_window, SESSION_HANDLE, session, uint32_t*, outgoing_window);
-    MOCKABLE_FUNCTION(, int, session_set_handle_max, SESSION_HANDLE, session, handle, handle_max);
-    MOCKABLE_FUNCTION(, int, session_get_handle_max, SESSION_HANDLE, session, handle*, handle_max);
```
dcristoloveanu: >MOCKABLE_FUNCTION [](start = 4, length = 17)

I don't quite like this split in internal and non internal for link/session/message.
I see no need for splitting things. Generally I do not like a layer of internals because it only complicates things. The link APIs are all available to its consumer and if a user decides to use the SDK at this link level, all power to them.

So please back it out. Pretty please.

dcristoloveanu: Same applies to session, message.

---
In reply to: [168280844](https://github.com/Azure/azure-uamqp-c/pull/212#discussion_r168280844) [](ancestors = 168280844)

inc/azure_uamqp_c/amqp_definitions.h
```diff
@@ -17,1016 +17,65 @@ extern "C" {
 #include "azure_uamqp_c/amqpvalue.h"
 #include "azure_c_shared_utility/umock_c_prod.h"
 
-/* role */
+#include "azure_uamqp_c/amqp_definitions_role.h"
+#include "azure_uamqp_c/amqp_definitions_sender_settle_mode.h"
```
dcristoloveanu: Very nice to have this header for whoever does not care to cherrypick.

inc/azure_uamqp_c/amqp_definitions_begin.h
```diff
@@ -0,0 +1,52 @@
+
+
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#ifndef AMQP_DEFINITIONS_BEGIN_H
```
dcristoloveanu: >P_DEFINITIONS [](start = 11, length = 13)

Consider:
It would be interesting maybe to have these generated types in a separate folder (probably very easy to change in the .cs that generates them.

And then we could improve the naming scheme a bit by having amqp_begin.h

So a user would include amqp_generated_types/amqp_begin.h for example.

I feel like it would be a cleaner presentation to the user.
This is a consider, but we could probably think a bit about it now. Anyway since most users are normally shielded by using the amqp_definitions.h main header, it should not be a big deal to change the naming scheme later also.

inc/azure_uamqp_c/session.h
dcristoloveanu: Since I'd like to see this backed out, I will wait for it.

---
In reply to: [168280940](https://github.com/Azure/azure-uamqp-c/pull/212#discussion_r168280940) [](ancestors = 168280940,168280844)

jspaith:
:)

---
In reply to: [365713457](https://github.com/Azure/azure-uamqp-c/pull/212#issuecomment-365713457) [](ancestors = 365713457)

inc/azure_uamqp_c/session.h
jspaith: Agreed per our discussion.  Next push will incorporate this back-out.

---
In reply to: [168282285](https://github.com/Azure/azure-uamqp-c/pull/212#discussion_r168282285) [](ancestors = 168282285,168280940,168280844)

PR #213, opened 2018-02-14
Mrzeropayload

markrad:
Thought the unit tests actually test the scenario when the payload is zero length they rely upon the fact that some implementations of malloc will return a valid pointer when a request for zero bytes is made. Unfortunately this is implementation specific and it is also valid to return a null pointer. On platforms that do that the AMQP message cannot be created with a zero length payload. This fix addresses that. No additional unit tests were added since this is already covered.


markrad:
@ewertons is added to the review. #Closed

PR #214, opened 2018-02-16
Fix error not being bubbled up by AMQP management

dcristoloveanu:


src/amqp_management.c
```diff
@@ -330,6 +331,46 @@ static AMQP_VALUE on_message_received(const void* context, MESSAGE_HANDLE messag
...
+            if (singlylinkedlist_remove(amqp_management->pending_operations, pending_operation_list_item_handle) != 0)
+            {
+                /* Tests_SRS_AMQP_MANAGEMENT_01_174: [ If any error occurs in removing the pending operation from the list `on_amqp_management_error` callback shall be invoked while passing the `on_amqp_management_error_context` as argument. ]*/
+                amqp_management->on_amqp_management_error(amqp_management->on_amqp_management_error_context);
+                LogError("Cannot remove pending operation");
+            }
+            else
+            {
+                /* Codes_SRS_AMQP_MANAGEMENT_01_173: [ - The callback associated with the pending operation shall be called with `AMQP_MANAGEMENT_EXECUTE_OPERATION_ERROR`. ]*/
+                pending_operation_message->on_execute_operation_complete(pending_operation_message->callback_context, AMQP_MANAGEMENT_EXECUTE_OPERATION_ERROR, 0, NULL);
```
annatisch: This doesn't build for me - as it seems to be missing a final NULL value on the end for the MESSAGE_HANDLE?

dcristoloveanu: Indeed.
I am very surprised at this. And ashamed. Sorry.

---
In reply to: [168890286](https://github.com/Azure/azure-uamqp-c/pull/214#discussion_r168890286) [](ancestors = 168890286)

dcristoloveanu:
Regarding the comment about the Rejected cases, it is the intended behavior as you mentioned. Basically we (client) are rejecting the response payload because we could not decode it properly. For example if we cannot read the correlation Id we would not know anyhow which outstanding AMQPMAN request to cancel.

It could be argued that for other cases we would know, but since we're rejecting the response payload, it can very well be that we will get a good payload afterwards.

An interesting feature here would be a timeout option on the amqp_management_execute_operation_async to allow the client to control the fact that the operation timed out and whatever correlated response would be received after the timeout would just be discarded.

Let me know if it makes sense. I also corrected the NULL issue you mentioned, so if this is fine for you I can merge it in.

annatisch:
LGTM! Thanks for the explanation :)

PR #220, opened 2018-03-05
Override default link credit

annatisch:


annatisch:
@dcristoloveanu - this is my first attempt at C, so let me know if this approach is too simplistic :)

dcristoloveanu:
This is exactly what I would expect! Unfortunately this module does not have unit tests, so I cannot ask you to add them :-).
I will merge it in.

---
In reply to: [370484734](https://github.com/Azure/azure-uamqp-c/pull/220#issuecomment-370484734) [](ancestors = 370484734)

PR #221, opened 2018-03-05
Exposed open performative properties for telemetry

annatisch:


src/connection.c
```diff
@@ -1542,6 +1564,61 @@ int connection_get_idle_timeout(CONNECTION_HANDLE connection, milliseconds* idle
     return result;
 }
 
+int connection_set_properties(CONNECTION_HANDLE connection, fields properties)
+{
+    int result;
+
+    /* Codes_SRS_CONNECTION_01_191: [If connection is NULL, connection_set_properties shall fail and return a non-zero value.] */
+    if (connection == NULL)
```
dcristoloveanu: This is a perfect sized change to exercise what we normally do for these SDKs regarding testing each module that is in.

We do 3 things for a module whenever we write it or we update it:
- Update the requirements document for the module (connection.md in this case). This requirement doc describes what the APIs of a module do, so that it can be tracked in the code (.c) and in the tests (_ut.c).
For this step we also use a VSCode plugin to easier generate unique requirement IDs, but it can also be done manually - manual is a bit more annoying, sorry about that.

-Update the unit tests (connection_ut.c) with the needed tests for the added/changed API.

- Update the code (.c).

I will create a PR to better describe this process in contribute.md so it is clearer.

This PR has only the code update, so it would be needed to do the other updates too.

Some of the modules in the repo (like link.c) do not have these requirements/tests done, so for those of course I do not expect this to happen as me or another team member would have to go and pay the technical debt.

But for modules that do have specs/tests we require this as part of the change so that we do not regress on coverage.

If needed also an E2E test can be added, although that can be a bit more involved.

dcristoloveanu: Let me know if you have issues while doing this and I can guide you though the changes. I would like this to go as smooth as we can for you so that subsequent changes get merged as fast as we can :-).

---
In reply to: [172395228](https://github.com/Azure/azure-uamqp-c/pull/221#discussion_r172395228) [](ancestors = 172395228)

annatisch:
Thanks @dcristoloveanu! Could we get this merged? :)

PR #223, opened 2018-03-09
Add LogError in link.c

dcristoloveanu:
Add error logging in link.c

### DIFF
--- a/src/link.c
+++ b/src/link.c
@@ -111,20 +111,31 @@ static int send_flow(LINK_INSTANCE* link)
 
     if (flow == NULL)
     {
+        LogError("NULL flow performative");
         result = __FAILURE__;
     }
     else
     {
-        if ((flow_set_link_credit(flow, link->current_link_credit) != 0) ||
-            (flow_set_handle(flow, link->handle) != 0) ||
-            (flow_set_delivery_count(flow, link->delivery_count) != 0))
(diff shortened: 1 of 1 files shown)


src/link.c
```diff
@@ -674,6 +738,7 @@ LINK_HANDLE link_create(SESSION_HANDLE session, const char* name, role role, AMQ
                 result->name = (char*)malloc(name_length + 1);
                 if (result->name == NULL)
                 {
+                    LogError("Cannot copy link name");
```
jspaith: Nit - consider "alloc" instead of "copy".

src/link.c
```diff
@@ -1233,6 +1365,7 @@ ASYNC_OPERATION_HANDLE link_transfer_async(LINK_HANDLE link, message_format mess
                             DELIVERY_INSTANCE* pending_delivery = GET_ASYNC_OPERATION_CONTEXT(DELIVERY_INSTANCE, result);
                             if (pending_delivery == NULL)
                             {
+                                LogError("Failed creating pending delivery");
```
jspaith: >creating [](start = 49, length = 8)

Is this a create?  Call looks more like a GET

dcristoloveanu: Thanks, it is a get. Changed.

---
In reply to: [173501836](https://github.com/Azure/azure-uamqp-c/pull/223#discussion_r173501836) [](ancestors = 173501836)

PR #224, opened 2018-03-14
Use strict XCode warnings

JetstreamRoySprowl:
Our compilers are letting several code issues through with no warning, including mismatched enum assignments. This change forces XCode to apply the stricter rules to CMake generated projects that it always applies to its wizard-generated iOS projects.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -149,6 +149,7 @@ add_library(uamqp
     ${uamqp_h_files}
     ${socketlistener_c_files}
     )
+setTargetBuildProperties(uamqp)
 
 target_link_libraries(uamqp aziotsharedutil)
 


CMakeLists.txt
```diff
@@ -149,6 +149,7 @@ add_library(uamqp
     ${uamqp_h_files}
     ${socketlistener_c_files}
     )
+setTargetBuildProperties(uamqp)
```
dcristoloveanu: >setTargetBuildProperties [](start = 0, length = 24)

If a warning is spewed by the compiler, then we should fix the warnings, rather than suppress them. Suppressing should be the last resort, only for the (unlikely) case that the compiler is wrong.

JetstreamRoySprowl: This change it the **opposite** of suppression; it makes the compiler _more_ strict, not less.

dcristoloveanu: Oh, sorry, my bad :-). By all means!

---
In reply to: [174605521](https://github.com/Azure/azure-uamqp-c/pull/224#discussion_r174605521) [](ancestors = 174605521)

dcristoloveanu: I started the build to merge it in.

JetstreamRoySprowl: Thanks!

PR #230, opened 2018-04-26
Add guards for the submodules so that git submodule update without recursive should be enough to build

dcristoloveanu:
Add guards for the submodules so that git submodule update without recursive should be enough to build


CMakeLists.txt
```diff
@@ -7,7 +7,9 @@ project(uamqp)
 FILE(READ ${CMAKE_CURRENT_LIST_DIR}/version.txt UAMQP_VERSION)
 
 # Include the common build rules for the C SDK
+if (NOT TARGET aziotsharedutil)
 include(deps/azure-c-shared-utility/configs/azure_iot_build_rules.cmake)
```
jebrando: Don't forget tab.

PR #240, opened 2018-05-31
Add support for extracting redirect information

dcristoloveanu:
This PR adds a handler for connection_close_received.
This allows extracting the error information once a connection is closed, thus allowing the redirect implementation by the user of the connection.
An E2E local TCP test was added. This prompted adding the info argument to connection_close.
I tinkered with the idea of making multiple event handlers for redirect, etc., but that seemed as overshoot for now.
This should solve at connection level the need to bubble up the redirect information.
Another PR will be posted for the link redirect information.

### DIFF
--- a/inc/azure_uamqp_c/connection.h
+++ b/inc/azure_uamqp_c/connection.h
@@ -12,6 +12,8 @@
 #include "azure_uamqp_c/amqp_frame_codec.h"
 #include "azure_uamqp_c/amqp_definitions_fields.h"
 #include "azure_uamqp_c/amqp_definitions_milliseconds.h"
+#include "azure_uamqp_c/amqp_definitions_error.h"
+#include "azure_uamqp_c/amqpvalue.h"
 
 #ifdef __cplusplus
 extern "C" {
@@ -19,6 +21,7 @@ extern "C" {
 
     typedef struct CONNECTION_INSTANCE_TAG* CONNECTION_HANDLE;
     typedef struct ENDPOINT_INSTANCE_TAG* ENDPOINT_HANDLE;
+    typedef struct ON_CONNECTION_CLOSED_EVENT_SUBSCRIPTION_TAG* ON_CONNECTION_CLOSED_EVENT_SUBSCRIPTION_HANDLE;
 
     typedef enum CONNECTION_STATE_TAG
     {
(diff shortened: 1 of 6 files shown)


src/connection.c
```diff
@@ -30,6 +30,12 @@ typedef enum RECEIVE_FRAME_STATE_TAG
     RECEIVE_FRAME_STATE_FRAME_DATA
 } RECEIVE_FRAME_STATE;
 
+typedef struct ON_CONNECTION_CLOSED_EVENT_SUBSCRIPTION_TAG
+{
+    ON_CONNECTION_CLOSE_RECEIVED on_connection_close_received;
+    void* context;
+} ON_CONNECTION_CLOSED_EVENT_SUBSCRIPTION;
```
jebrando: >ON_CONNECTION_CLOSED_EVENT_SUBSCRIPTION [](start = 2, length = 39)

Curious naming, I thought this was a function point, but it is a structure.

inc/azure_uamqp_c/connection.h
```diff
@@ -12,13 +12,16 @@
...
+#include "azure_uamqp_c/amqp_definitions_error.h"
+#include "azure_uamqp_c/amqpvalue.h"
 
 #ifdef __cplusplus
 extern "C" {
 #endif /* __cplusplus */
 
     typedef struct CONNECTION_INSTANCE_TAG* CONNECTION_HANDLE;
     typedef struct ENDPOINT_INSTANCE_TAG* ENDPOINT_HANDLE;
+    typedef void* EVENT_SUBSCRIPTION_HANDLE;
```
jebrando: >void* [](start = 12, length = 5)

Should this be a ON_CONNECTION_CLOSED_EVENT_SUBSCRIPTION_TAG*?

dcristoloveanu: Thank you Sir. I just realized that my initial idea of having a generic event subscription handle does not really help, it actually takes away the compiler checks

---
In reply to: [192261129](https://github.com/Azure/azure-uamqp-c/pull/240#discussion_r192261129) [](ancestors = 192261129)

dcristoloveanu: So, yes, this will be ON_CONNECTION_CLOSED_EVENT_SUBSCRIPTION_TAG

---
In reply to: [192285930](https://github.com/Azure/azure-uamqp-c/pull/240#discussion_r192285930) [](ancestors = 192285930,192261129)

src/connection.c
dcristoloveanu: Yeah, /poetry. Noted, though about it, but decided to leave as is :-)

---
In reply to: [192261066](https://github.com/Azure/azure-uamqp-c/pull/240#discussion_r192261066) [](ancestors = 192261066)

PR #241, opened 2018-06-05
 Add link detach received handler to be used for redirect

dcristoloveanu:
Add link detach received handler to be used for redirect


dcristoloveanu:
There are currently some issues with the gated system that is being used that prevents from merging changes. There is someone looking at this and as soon as the issues are fixed this will get merged.

Thanks for the review!
/Dan

Issue #251, opened 2018-07-11
Proposal for exposing send disposition delivery state


annatisch:
Hi there,

I would like to surface the ["state" field of the disposition performative](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#type-disposition) for the message sender, and pass this value into the `on_message_send_complete` callback.

When running high performance tests against EventHubs I sometimes get the following disposition:
```
'<- [DISPOSITION]* {true,721,NULL,true,* {* {com.microsoft:server-busy,The request was terminated because the entity is being throttled. Error code : 50002. Please wait 4 seconds and try again. Reference:eff2f1cf-d942-4f77-ac6b-3c3842edd98b, TrackingId:37f5c5020000e8fd00009c6c5b467515_G9_B2, SystemTracker:int7bn3004-private-python-0-5aedf-2:eventhub:pytest128~1279, Timestamp:7/11/2018 9:22:54 PM,NULL}},NULL}'
```

Currently I have a blanket retry policy on message send failures - but it would be useful to actually be able to apply this policy to only retry-able errors and take other actions as needed (for example, in this case, to pause the client for the recommended 4 seconds).

I did a quick prototype which I am currently testing with that looks something like this:

**message_sender.h**
```
typedef void(*ON_MESSAGE_SEND_COMPLETE)(void* context, MESSAGE_SEND_RESULT send_result, AMQP_VALUE delivery_state);
```
**message_sender.c**
```
static void on_delivery_settled(void* context, delivery_number delivery_no, LINK_DELIVERY_SETTLE_REASON reason, AMQP_VALUE delivery_state)
{
    ASYNC_OPERATION_HANDLE pending_send = (ASYNC_OPERATION_HANDLE)context;
    MESSAGE_WITH_CALLBACK* message_with_callback = GET_ASYNC_OPERATION_CONTEXT(MESSAGE_WITH_CALLBACK, pending_send);
    MESSAGE_SENDER_INSTANCE* message_sender = (MESSAGE_SENDER_INSTANCE*)message_with_callback->message_sender;
    (void)delivery_no;

    if (message_with_callback->on_message_send_complete != NULL)
    {
        switch (reason)
        {
        case LINK_DELIVERY_SETTLE_REASON_DISPOSITION_RECEIVED:
            if (delivery_state == NULL)
            {
                LogError("delivery state not provided");
            }
            else
            {
[...]

PR #253, opened 2018-07-20
Surface message disposition delivery-state

annatisch:
Here's my changes according to issue #251 

In addition:
- I've made modifications to link.c so that the subscribe_to_detach_event callback is called regardless of whether there's an error in the DETACH frame. I did this because I've noticed some services (i.e. EventHubs) will randomly send an empty DETACH frame for no reason - clients are supposed to handle this and reconnect - which is easier to do when I can capture the event in the callback. Open to your thoughts on this though :)
- I have not done anything to make use of the new delivery-state fields in amqp_management.c yet. I do intend to revisit this, as well as DETACH events in the management links to further capture and surface errors here.

### DIFF
--- a/devdoc/amqp_management_requirements.md
+++ b/devdoc/amqp_management_requirements.md
@@ -309,7 +309,7 @@ AMQP_VALUE on_message_received(const void* context, MESSAGE_HANDLE message)
 ### on_message_send_complete
 
 ```c
-void on_message_send_complete(void* context, MESSAGE_SEND_RESULT send_result)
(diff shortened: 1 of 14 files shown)


annatisch:
ping @dcristoloveanu and @jebrando for review :)

dcristoloveanu:
Sorry for the delay, I just got to this, looks very clean :-).

---
In reply to: [407864450](https://github.com/Azure/azure-uamqp-c/pull/253#issuecomment-407864450) [](ancestors = 407864450)

PR #254, opened 2018-08-02
Include header files outside the `extern "C"` (fix for azure-iot-sdk-c#577)

BigPeteB:
This fixes errors when compiling as C++ code, reported in Azure/azure-iot-sdk-c#577, q.v.


dcristoloveanu:
Thanks for the fix, merging in.

PR #255, opened 2018-08-06
Move indication of error'd messages in message_sender to work around …

dcristoloveanu:
…bad consumers that would pull the rug from under message_sender

### DIFF
--- a/src/message_sender.c
+++ b/src/message_sender.c
@@ -690,9 +690,9 @@ static void on_link_state_changed(void* context, LINK_STATE new_link_state, LINK
         if ((message_sender->message_sender_state == MESSAGE_SENDER_STATE_OPEN) ||
             (message_sender->message_sender_state == MESSAGE_SENDER_STATE_CLOSING))
         {
-            /* User initiated transition, we should be good */
-            set_message_sender_state(message_sender, MESSAGE_SENDER_STATE_IDLE);
+            /* switch to closing so that no more requests should be accepted */
             indicate_all_messages_as_error(message_sender);
+            set_message_sender_state(message_sender, MESSAGE_SENDER_STATE_IDLE);
         }
         else if (message_sender->message_sender_state != MESSAGE_SENDER_STATE_IDLE)
         {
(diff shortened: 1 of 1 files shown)


src/message_sender.c
```diff
@@ -690,9 +690,9 @@ static void on_link_state_changed(void* context, LINK_STATE new_link_state, LINK
         if ((message_sender->message_sender_state == MESSAGE_SENDER_STATE_OPEN) ||
             (message_sender->message_sender_state == MESSAGE_SENDER_STATE_CLOSING))
         {
-            /* User initiated transition, we should be good */
-            set_message_sender_state(message_sender, MESSAGE_SENDER_STATE_IDLE);
+            /* switch to closing so that no more requests should be accepted */
```
ewertons: I believe we once changed this because it would give more information to the upper layer when messages start to be returned back as failed.

What is the reason for changing this again?

src/message_sender.c
```diff
@@ -703,8 +703,8 @@ static void on_link_state_changed(void* context, LINK_STATE new_link_state, LINK
     case LINK_STATE_ERROR:
         if (message_sender->message_sender_state != MESSAGE_SENDER_STATE_ERROR)
         {
-            set_message_sender_state(message_sender, MESSAGE_SENDER_STATE_ERROR);
             indicate_all_messages_as_error(message_sender);
```
ewertons: Same here.

dcristoloveanu: Well, indeed, I do remember discussing these lines.
So here goes the whole thought process with all the ramifications.

Handling of not transmitted messages when message sender closes:
In any object that has open/close, when close is called, normally what should happen is this:
- Set the state to closing.
- Indicate all pending requests as BECAUSE_CLOSE/ERROR.
- Maybe do other stuff
- Set the state to closed
The reason indication of the pending messages with error should happen after setting the state to CLOSING would be to avoid that the consumer erroneously requeues forever the same requests.
Setting the state to closing would allow checking it in all new calls to send_async, and would effectively bounce back all these requeued requests.
Why I did not do that now: I do not know how the IoT SDK would react to the CLOSING state and I did not want to induce this pain on the IoT SDK right now.
Also notice that there are other places that have a similar pattern of first indicating the messages with ERROR and then indicating to the user that the state has changed.

Why the change:
IoT SDK is destroying the message sender in the state changed callback (I believe the D2C or C2D methods portion in the transport does that).
That is actually a no-no and it leads to an AV. Obviously it should be fixed, as the consumer should not pull the rug from under the message_sender.
I believe that is a bigger endeavor and it this should be tracked with an Issue in the IoT SDK and fixed.

Until then this workaround in uamqp should alleviate the issues Mark is seing without much disturbance.

HTH,
/Dan

PR #256, opened 2018-08-21
Fix clang static analysis errors

ewertons:
These are part of the failures listed by clang.

### DIFF
--- a/src/amqp_management.c
+++ b/src/amqp_management.c
@@ -218,11 +218,10 @@ static AMQP_VALUE on_message_received(const void* context, MESSAGE_HANDLE messag
                                                 OPERATION_MESSAGE_INSTANCE* operation_message = (OPERATION_MESSAGE_INSTANCE*)singlylinkedlist_item_get_value(list_item_handle);
                                                 if (operation_message == NULL)
                                                 {
-                                                    /* Codes_SRS_AMQP_MANAGEMENT_01_117: [ If iterating through the pending operations list fails, an error shall be indicated by calling `on_amqp_management_error` and passing the `on_amqp_management_error_context` to it. ]*/
                                                     LogError("Could not create status-description amqp value");
-                                                    amqp_management->on_amqp_management_error(amqp_management->on_amqp_management_error_context);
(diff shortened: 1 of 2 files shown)


src/amqpvalue.c
```diff
@@ -4963,7 +4963,6 @@ static int internal_decoder_decode_bytes(INTERNAL_DECODER_DATA* internal_decoder
                             if (list_item == NULL)
                             {
                                 internal_decoder_data->decoder_state = DECODER_STATE_ERROR;
-                                result = __FAILURE__;
```
dcristoloveanu: >result = __FAILURE__; [](start = 32, length = 21)

Looking at this I actually think we should return failure in these cases. One easy way would be in ln 4980 an extra check for internal_decoder_data->decoder_state == DECODER_STATE_ERROR.

Something like:
if (
  (internal_decoder_data->inner_decoder == NULL) ||
  (internal_decoder_data->decoder_state == DECODER_STATE_ERROR)
)

Same applies for the places below.

src/amqp_management.c
```diff
@@ -218,11 +218,10 @@ static AMQP_VALUE on_message_received(const void* context, MESSAGE_HANDLE messag
...
                                                     LogError("Could not create status-description amqp value");
-                                                    amqp_management->on_amqp_management_error(amqp_management->on_amqp_management_error_context);
+                                                    /* Codes_SRS_AMQP_MANAGEMENT_01_117: [ If iterating through the pending operations list fails, an error shall be indicated by calling `on_amqp_management_error` and passing the `on_amqp_management_error_context` to it. ]*/
                                                     /* Codes_SRS_AMQP_MANAGEMENT_01_135: [ When an error occurs in creating AMQP values (for status code, etc.) `on_message_received` shall call `messaging_delivery_released` and return the created delivery AMQP value. ]*/
-                                                    result = messaging_delivery_released();
+                                                    is_error = true;
```
dcristoloveanu: >s_error = true; [](start = 53, length = 15)

Neat!

dcristoloveanu:
This is currently clutter. If you plan to fix this please reopen the PR.

PR #257, opened 2018-08-24
Fix Clang warning: Undefined allocation of 0 bytes

zolvarga:


src/message_sender.c
```diff
@@ -361,38 +361,46 @@ static SEND_ONE_MESSAGE_RESULT send_one_message(MESSAGE_SENDER_INSTANCE* message
                 }
                 else
                 {
-                    for (i = 0; i < body_data_count; i++)
+                    if (body_data_count <= 0)
```
dcristoloveanu: >< [](start = 40, length = 1)

body_data_count is size_t. It cannot be negative per C standard.

zolvarga: Updated the PR...

PR #258, opened 2018-09-03
MOD:%zu-%lu to mbed related files

SeanZhang0920:
As %zu cannot be shown correctly on mbed os5, I modified %zu to %lu in some mbed-os5-related files.

### DIFF
--- a/tests/amqp_frame_codec_ut/amqp_frame_codec_ut.c
+++ b/tests/amqp_frame_codec_ut/amqp_frame_codec_ut.c
@@ -993,7 +993,7 @@ TEST_FUNCTION(amqp_performatives_are_encoded_successfully)
 
         // assert
         ASSERT_ARE_EQUAL(int, 0, result);
-        (void)sprintf(test_string, "test %zu", i);
+        (void)sprintf(test_string, "test %lu%lu", i);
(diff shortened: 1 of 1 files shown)


tests/amqp_frame_codec_ut/amqp_frame_codec_ut.c
```diff
@@ -993,7 +993,7 @@ TEST_FUNCTION(amqp_performatives_are_encoded_successfully)
 
         // assert
         ASSERT_ARE_EQUAL(int, 0, result);
-        (void)sprintf(test_string, "test %zu", i);
+        (void)sprintf(test_string, "test %lu%lu", i);
```
dcristoloveanu: >%lu%lu [](start = 41, length = 6)

This looks wrong, should be "test %lu", right?

dcristoloveanu: >i [](start = 50, length = 1)

Could you please cast this to (unsigned long) to be explicit about it?
Thanks!

SeanZhang0920: zu
when using %zu, message cannot be shown in correct format on the platform of mbed os5. As I modified it to %lu, the message can correctly shown as below.
lu
And sorry about the "%lu%lu" mistake, I have corrected it and pushed my modification.

Best Regards

PR #261, opened 2018-10-09
Added attach callback for existing link endpoint

annatisch:
I feel that the on_link_attach callback should be called regardless of the whether the link endpoint is newly created, or was created earlier.

I am not entirely sure I have the error scenario correct in line 512 (i.e. whether I need to do more than just setting the state to DETACHING).

Thanks,
Anna


annatisch:
ping @dcristoloveanu :)

annatisch:
Added this patch to PR #271

Issue #269, opened 2018-10-22
Array encoding difficulties


annatisch:
Hi @dcristoloveanu - I'm at a bit of a loss and hoping you can help me out here.

So I'm looking at implementing two commands for service bus - both of which deal in arrays.

The first - [Schedule messages](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-amqp-request-response#schedule-message) and [cancel scheduled messages](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-amqp-request-response#cancel-scheduled-message).

The data flow, in short, is that the response body for the schedule operation will be an AMQP Array of Longs. This is then fed directly into the cancel operation.

However, when I send an Array containing a single long to the service - the service seems to decode it as a different value than that which I sent:
```
AMQP array of sequence numbers to cancel: [69805794325242792]
Error from service: Messages not found. Missing sequence numbers: 101000104

Array as encoded by Service Bus: e00a018100f80000060523a8
Array as encoded by client: e00a018100f80000060523a8
```

If I try the operation using two or more values in the array I get a decode error:
```
AMQP array of sequence numbers to cancel: [40250921740623951, 36028797089964112]
Error from service: The format code '0x50' at frame buffer offset '179' is invalid or unexpected

Array as encoded by Service Bus: e0130281008f0000043b604f8100800000043b6050
Array as encoded by client: e0130281008f0000043b604f8100800000043b6050
```

The second operation is [renew message lock](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-amqp-request-response#message-renew-lock).

The data flow in this case is one-way, sending an Array of UUIDs.
Weidly, if I attempt to renew a single lock token, it works:
```
AMQP array of UUID: [ebe26a2f-7c6a-4987-9943-fa04da810029]
Response from the service: {'expirations': [1540248072441]}

Array as encoded by client: e0120198ebe26a2f7c6a49879943fa04da810029
```

But if I attempt to renew 2 or more values in the array, again the decode error:
```
AMQP array of UUID: [4e8c1d72-5b09-4a65-a57c-554edc7a0029, 46fca782-d8cc-439c-a7e9-7d3d0b0d0037]
Error from the service: The format code '0x37' at frame buffer offset '175' is invalid or unexpected

Array as encoded by client: e02302984e8c1d725b094a65a57c554edc7a00299846fca782d8cc439ca7e97d3d0b0d0037
```

[...]

PR #271, opened 2018-11-01
Fix for array encoding + link ATTACH callback

annatisch:
Fix for issue #269 

I have changed the type encoding functions to separate out the constructor encoding and the data encoding into separate methods for each type. This allows me to call them separately to encode the elements in the array without duplicating the encoding logic already used for generic type encoding.

It is however, quite verbose - so happy to hear feedback on alternative implementations.

In terms of supported types with which to populate an array - my fix only supports basic types, not complex types like list or map (though this could probably be added later if needed):
- NULL
- boolean
- ubyte / byte
- ushort / short
- uint / int
- ulong / long
- float
- double
- timestamp
- uuid
- binary
- string
- symbol

As mentioned in the issue thread - I've opted to only use the largest available type encoding to prevent the need to scan the array upfront to select the appropriate constructor. This could be amended later to improve data efficiency.

One area that may need further review is the size calculation. Given that subsequent constructor bytes are no longer included, they should be subtracted from the size.

### DIFF
--- a/inc/azure_uamqp_c/amqpvalue.h
+++ b/inc/azure_uamqp_c/amqpvalue.h
@@ -92,8 +92,6 @@ extern "C" {
 
     MOCKABLE_FUNCTION(, int, amqpvalue_encode, AMQP_VALUE, value, AMQPVALUE_ENCODER_OUTPUT, encoder_output, void*, context);
     MOCKABLE_FUNCTION(, int, amqpvalue_get_encoded_size, AMQP_VALUE, value, size_t*, encoded_size);
-    MOCKABLE_FUNCTION(, int, amqpvalue_encode_array_item, AMQP_VALUE, item, bool, first_element, AMQPVALUE_ENCODER_OUTPUT, encoder_output, void*, context);
-    MOCKABLE_FUNCTION(, int, amqpvalue_get_encoded_array_item_size, AMQP_VALUE, item, size_t*, encoded_size);
 
     /* decoding */
     typedef struct AMQPVALUE_DECODER_HANDLE_DATA_TAG* AMQPVALUE_DECODER_HANDLE;
--- a/inc/azure_uamqp_c/amqpvalue.h
+++ b/inc/azure_uamqp_c/amqpvalue.h
@@ -92,6 +92,8 @@ extern "C" {
 
     MOCKABLE_FUNCTION(, int, amqpvalue_encode, AMQP_VALUE, value, AMQPVALUE_ENCODER_OUTPUT, encoder_output, void*, context);
     MOCKABLE_FUNCTION(, int, amqpvalue_get_encoded_size, AMQP_VALUE, value, size_t*, encoded_size);
+    MOCKABLE_FUNCTION(, int, amqpvalue_encode_array_item, AMQP_VALUE, item, bool, first_element, AMQPVALUE_ENCODER_OUTPUT, encoder_output, void*, context);
+    MOCKABLE_FUNCTION(, int, amqpvalue_get_encoded_array_item_size, AMQP_VALUE, item, size_t*, encoded_size);
 
     /* decoding */
     typedef struct AMQPVALUE_DECODER_HANDLE_DATA_TAG* AMQPVALUE_DECODER_HANDLE;
--- a/src/amqpvalue.c
+++ b/src/amqpvalue.c
@@ -2268,6 +2268,62 @@ static int output_bytes(AMQPVALUE_ENCODER_OUTPUT encoder_output, void* context, 
     return result;
 }
 
+static int encode_boolean_constructor(AMQPVALUE_ENCODER_OUTPUT encoder_output, void* context)
+{
+    int result;
+
+    /* Codes_SRS_AMQPVALUE_01_270: [<encoding code="0x56" category="fixed" width="1" label="boolean with the octet 0x00 being false and octet 0x01 being true"/>] */
+    if (output_byte(encoder_output, context, 0x56) != 0)
+    {
+        /* Codes_SRS_AMQPVALUE_01_274: [When the encoder output function fails, amqpvalue_encode shall fail and return a non-zero value.] */
+        LogError("Failed encoding boolean constructor");
+        result = __FAILURE__;
+    }
+    else
+    {
+        /* Codes_SRS_AMQPVALUE_01_266: [On success amqpvalue_encode shall return 0.] */
+        result = 0;
+    }
+
+    return result;
+}
+
+static int encode_boolean_value(AMQPVALUE_ENCODER_OUTPUT encoder_output, void* context, bool value)
+{
+    int result;
+
+    if (value == false)
+    {
+        /* Codes_SRS_AMQPVALUE_01_270: [<encoding code="0x56" category="fixed" width="1" label="boolean with the octet 0x00 being false and octet 0x01 being true"/>] */
+        if (output_byte(encoder_output, context, 0x00) != 0)
+        {
+            /* Codes_SRS_AMQPVALUE_01_274: [When the encoder output function fails, amqpvalue_encode shall fail and return a non-zero value.] */
+            LogError("Failed encoding boolean value");
+            result = __FAILURE__;
+        }
+        else
+        {
+            result = 0;
+        }
+    }
+    else
+    {
+        /* Codes_SRS_AMQPVALUE_01_270: [<encoding code="0x56" category="fixed" width="1" label="boolean with the octet 0x00 being false and octet 0x01 being true"/>] */
+        if (output_byte(encoder_output, context, 0x01) != 0)
+        {
+            /* Codes_SRS_AMQPVALUE_01_274: [When the encoder output function fails, amqpvalue_encode shall fail and return a non-zero value.] */
+            LogError("Failed encoding boolean value");
+            result = __FAILURE__;
+        }
+        else
+        {
+            result = 0;
+        }
+    }
+
+    return result;
+}
+
 static int encode_boolean(AMQPVALUE_ENCODER_OUTPUT encoder_output, void* context, bool value)
 {
     int result;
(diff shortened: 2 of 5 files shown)


annatisch:
Combined with PR #261

src/amqpvalue.c
```diff
@@ -3183,22 +3960,246 @@ static int encode_array(AMQPVALUE_ENCODER_OUTPUT encoder_output, void* context,
 
         if (result == 0)
         {
+            AMQP_TYPE array_type = AMQP_TYPE_UNKNOWN;
+
             for (i = 0; i < count; i++)
             {
-                if (amqpvalue_encode(items[i], encoder_output, context) != 0)
+                AMQP_VALUE_DATA* value_data = (AMQP_VALUE_DATA*)items[i];
+                if ((array_type != AMQP_TYPE_UNKNOWN) && (value_data->type != array_type))
```
dcristoloveanu: >if ((array_type != AMQP_TYPE_UNKNOWN) && (value_data->type != array_type)) [](start = 16, length = 74)

This should be caught at the time of creating the array.

src/amqpvalue.c
```diff
@@ -3183,22 +3960,246 @@ static int encode_array(AMQPVALUE_ENCODER_OUTPUT encoder_output, void* context,
...
                 {
+                    LogError("Item type '%d' inconsistent with array type '%d'", (int)value_data->type, (int)array_type);
```
dcristoloveanu: >(int)value_data->type [](start = 81, length = 21)

Consider using ENUM_TO_STRING macro to print the nice stringified value in the enum.
uAMQP was written some time ago and I think not all of it uses it.

That would require having the enum values defined like 

#define AMQP_VALUE_TYPE_VALUES \
  AMQP_UNKNOWN, \
  ...

and then using the DEFINE_ENUM macro to actually have the enum defined and then DEFINE_ENUM_STRINGS to have a char* array with the stringified version of the values.

A consider, if it seems too much skip it please :-). Just thought of mentioning that that is how we would write it today :-).

dcristoloveanu:
>         if (amqpvalue_get_encoded_size(items[i], &item_size) != 0)

Constructor is always 1 byte

---
Refers to: src/amqpvalue.c:3888 in 9088bc6. [](commit_id = 9088bc66950afce79f2f54bcea63c7661de2610e, deletion_comment = False)

dcristoloveanu:
> static int encode_map(AMQPVALUE_ENCODER_OUTPUT encoder_output, void* context, uint32_t count, AMQP_MAP_KEY_VALUE_PAIR* pairs)

This should also be encodeable in an array

---
Refers to: src/amqpvalue.c:3748 in 9088bc6. [](commit_id = 9088bc66950afce79f2f54bcea63c7661de2610e, deletion_comment = False)

dcristoloveanu:
> static int encode_map(AMQPVALUE_ENCODER_OUTPUT encoder_output, void* context, uint32_t count, AMQP_MAP_KEY_VALUE_PAIR* pairs)

Basically all types should be encodeable in an array

---
In reply to: [435281473](https://github.com/Azure/azure-uamqp-c/pull/271#issuecomment-435281473) [](ancestors = 435281473)

---
Refers to: src/amqpvalue.c:3748 in 9088bc6. [](commit_id = 9088bc66950afce79f2f54bcea63c7661de2610e, deletion_comment = False)

dcristoloveanu:
> static int encode_map(AMQPVALUE_ENCODER_OUTPUT encoder_output, void* context, uint32_t count, AMQP_MAP_KEY_VALUE_PAIR* pairs)

Including arrays :-)

---
In reply to: [435281520](https://github.com/Azure/azure-uamqp-c/pull/271#issuecomment-435281520) [](ancestors = 435281520,435281473)

---
Refers to: src/amqpvalue.c:3748 in 9088bc6. [](commit_id = 9088bc66950afce79f2f54bcea63c7661de2610e, deletion_comment = False)

dcristoloveanu:
>         if (amqpvalue_get_encoded_size(items[i], &item_size) != 0)

This seems wrong (I think you mentioned it in your PR).
The easy way out though is to substract 1 from the item_size as the constructor is always 1 byte per the standard.

So the sum of (encoded_size - 1) and then adding 1 for the constructor that appears only once should solve the issue.

---
Refers to: src/amqpvalue.c:3888 in 9088bc6. [](commit_id = 9088bc66950afce79f2f54bcea63c7661de2610e, deletion_comment = False)

dcristoloveanu:
>         if (amqpvalue_get_encoded_size(items[i], &item_size) != 0)

Waiting for this and the map and other types to be encodeable in the array (they have to have thje same separation of encode_constructor/encode_value I presume)

---
In reply to: [435281897](https://github.com/Azure/azure-uamqp-c/pull/271#issuecomment-435281897) [](ancestors = 435281897)

---
Refers to: src/amqpvalue.c:3888 in 9088bc6. [](commit_id = 9088bc66950afce79f2f54bcea63c7661de2610e, deletion_comment = False)

src/amqpvalue.c
annatisch: Thanks, I've refactored this out.

src/amqpvalue.c
annatisch: I've refactored out the array type check - so I will leave this for now.

annatisch:
Thanks for the review @dcristoloveanu!

I think I have addressed all the feedback. I'm pretty confident in the array size calculation now - turns out subtracting 1 isn't sufficient as: 
- it doesn't account for single byte items like NULL
- the amqpvalue_get_encoded_size function doesn't take into account that we always use the largest available encoding for an array item.

So I've refactored the encode array item value and calculating its size into separate functions.
I also added support for arrays of arrays, lists and maps :)

inc/azure_uamqp_c/amqpvalue.h
```diff
@@ -92,6 +92,8 @@ extern "C" {
 
     MOCKABLE_FUNCTION(, int, amqpvalue_encode, AMQP_VALUE, value, AMQPVALUE_ENCODER_OUTPUT, encoder_output, void*, context);
     MOCKABLE_FUNCTION(, int, amqpvalue_get_encoded_size, AMQP_VALUE, value, size_t*, encoded_size);
+    MOCKABLE_FUNCTION(, int, amqpvalue_encode_array_item, AMQP_VALUE, item, bool, first_element, AMQPVALUE_ENCODER_OUTPUT, encoder_output, void*, context);
```
dcristoloveanu: >amqpvalue_encode_array_item [](start = 29, length = 27)

I think these are internal to the amqpvalue module.
A user of the module would always encode an entire array, so they should be removed from the .h and made static.

dcristoloveanu: I mean both of them: amqpvalue_encode_array_item and amqpvalue_get_encoded_array_item_size

---
In reply to: [230628469](https://github.com/Azure/azure-uamqp-c/pull/271#discussion_r230628469) [](ancestors = 230628469)

dcristoloveanu:
> int amqpvalue_get_encoded_array_item_size(AMQP_VALUE item, size_t* encoded_size)

static

---
Refers to: src/amqpvalue.c:4696 in 1f8c25f. [](commit_id = 1f8c25f04a86cf29156466bb8f1d38007c256933, deletion_comment = False)

dcristoloveanu:
> int amqpvalue_encode_array_item(AMQP_VALUE item, bool first_element, AMQPVALUE_ENCODER_OUTPUT encoder_output, void* context)

static

---
Refers to: src/amqpvalue.c:4434 in 1f8c25f. [](commit_id = 1f8c25f04a86cf29156466bb8f1d38007c256933, deletion_comment = False)

dcristoloveanu:
> static int encode_array_constructor(AMQPVALUE_ENCODER_OUTPUT encoder_output, void* context, bool use_smallest)

Cool, arrays work now :-)

---
Refers to: src/amqpvalue.c:4073 in 1f8c25f. [](commit_id = 1f8c25f04a86cf29156466bb8f1d38007c256933, deletion_comment = False)

dcristoloveanu:
Having a look again ...

---
In reply to: [435598905](https://github.com/Azure/azure-uamqp-c/pull/271#issuecomment-435598905) [](ancestors = 435598905)

dcristoloveanu:
>                 result = output_byte(encoder_output, context, (unsigned char)0x40);

I do not think it is correct to have the NULL type specially treated.
For an array of NULL, the way I read the AMQP spec, we'd have a NULL constructor (0x40) and that is basically it, no bytes for the items in the array.

So I really think there should be an encode_NULL_constructor and encode_NULL_value for consistency

The way this is coded here it would produce a 0x40 0x40 for an array of 2 items.

And that would fail decoding.

---
Refers to: src/amqpvalue.c:4460 in 1f8c25f. [](commit_id = 1f8c25f04a86cf29156466bb8f1d38007c256933, deletion_comment = False)

dcristoloveanu:
"- it doesn't account for single byte items like NULL"
I believe that it should. I have another comment in the review saying that encoding a NULL should be similar to other types.
Essentially we should encode the constructor for NULL (0x40) for the first item and then each encode_NULL_value should add no extra bytes.
So I really think we should normalize this for NULL as otherwise there'd be other decoding failures that will pop up.

"- the amqpvalue_get_encoded_size function doesn't take into account that we always use the largest available encoding for an array item."
Ok, I understand this :-), so the current code specifically for array items should work. The only comment here is to remove the extra functions from the .h as they are actually meant to be internal.

---
In reply to: [435747610](https://github.com/Azure/azure-uamqp-c/pull/271#issuecomment-435747610) [](ancestors = 435747610,435598905)

dcristoloveanu:
Waiting for these, but I think once these are done you should be set!

---
In reply to: [435750045](https://github.com/Azure/azure-uamqp-c/pull/271#issuecomment-435750045) [](ancestors = 435750045,435747610,435598905)

annatisch:
Thanks @dcristoloveanu!

That's very interesting your comment on NULL handling - I didn't realize that in a "single-byte" encoded value was considered a constructor only, and therefore not duplicated for the rest of the array.
In that case it would seem that all the existing tests for [encoding ](https://github.com/Azure/azure-uamqp-c/blob/master/tests/amqpvalue_ut/amqpvalue_ut.c#L9324)and [decoding ](https://github.com/Azure/azure-uamqp-c/blob/master/tests/amqpvalue_ut/amqpvalue_ut.c#L16153)an array of NULL values are currently incorrect. I will get that fixed!

Regarding my inclusion of amqpvalue_encode_array_item and amqpvalue_get_encoded_array_item_size in the header file - the reason I did this is because I can't get the project to build otherwise... because the various array encoding functions call into each other, I am unable to order them such that I don't get at least one "function is undefined" warning. What is the standard way of getting around this?

dcristoloveanu:
No worries.

For the amqtvalue_encode_array_item and amqpvalue_get_encoded_array_item_size the standard thing to do is have the function declaration at the top of the translation unit without the function body.
It is called a forward declaration:

int amqpvalue_encode_array_item(AMQP_VALUE item, bool first_element, AMQPVALUE_ENCODER_OUTPUT encoder_output, void* context);

---
In reply to: [435937840](https://github.com/Azure/azure-uamqp-c/pull/271#issuecomment-435937840) [](ancestors = 435937840)

annatisch:
Thanks! I added the forward declarations and it all builds correctly :)

So after I adjusted the array_decode tests to use only a single NULL byte - it seems that the original decode_array logic does not work this way - and it anticipates the number of NULL bytes to match the count value. It looks like some fixes will be needed here as well - is this what you would expect?

dcristoloveanu:
Regarding the NULL arrays: yes. The code was wrong there based on the spec. I just confirmed this with Xin who implemented the C# library. We both agree that the AMQP spec here is rather strange and does not allow for very good use of arrays, but it is what it is and we should conform to the spec.

So, if you don't mind doing that change it would be best to have it conformant.

---
In reply to: [436000668](https://github.com/Azure/azure-uamqp-c/pull/271#issuecomment-436000668) [](ancestors = 436000668)

annatisch:
Thanks for the confirmation @dcristoloveanu!

I added [this check here](https://github.com/Azure/azure-uamqp-c/pull/271/files#diff-a2ca541f4fdc2a7529cc4b878e32000eR6814) so that if the last decoded item was only a single byte, there's no remaining size allocated, and yet the item count has not been reached - then don't increment the buffer so that the same byte can be decoded again. It seems to work with the updated tests.

Also addressed the other feedback.

annatisch:
@dcristoloveanu - I was just chatting with a maintainer for the NodeJS AMQP lib, and it seems when they encode NULL arrays, their size counter still counts the bytes that aren't there.
For example, an array of 4 NULL values would encode like:
E0 05 04 40

Where as my encoding/decoding currently assumes:
E0 01 04 40

What are your thoughts on this?

dcristoloveanu:
Well that is incorrect per the AMQP standard. Probably it will not matter as encoding an array of NULL values is not very common (or useful). But from the point of view of the spec it still is incorrect and she/he might want to fix it :-).

---
In reply to: [436335706](https://github.com/Azure/azure-uamqp-c/pull/271#issuecomment-436335706) [](ancestors = 436335706)

annatisch:
Thanks for the review @dcristoloveanu - however there seems to be something in this patch that is causing issues in my code - **please don't merge just yet** while I try to debug it.

annatisch:
So it seems that I made some incorrect assumptions about the values of `size` and `inner_used_bytes`.
This was causing fails in SASL auth, the frame of which apparently uses an array of symbols where the size calculation was causing issues.

I've altered my logic to explicitly not increment the buffer only if the item constructor byte is one of the 5 `width="0"` types defined in the spec. It looks a bit ugly, but should be more reliable then depending on size/decoded bytes.....

My scenario tests are now passing - so if this looks okay to you, feel free to merge :)

dcristoloveanu:
Hi Anna,

Actually 5 of the UTs of amqpvaue_ut were failing also for the encoding cases.
The problem was that the array item sizes were not incremented properly (there was a *encoded_size++ instead of (*encoded_size)++;).
This was actually caught by the OsX gate that runs clang and clang flagged with a warning this offending line. Once I fixed this I noticed that the UT expectations were off regarding the size of the bytes.

To clarify, an array is made of:
- size - n bytes that follow after the size component (including count, constructor, etc.)
- count - no of items in the array
- constructor - 1 byte
- items data

size was not taking into account the complete thing in some of the tests, so I fixed that and am trying to merge in with this branch: annatisch-array-fix
If you find anything else that you need to correct on this PR or want to try those changes on your specific case, can you please merge annatisch-array-fix into your branch and then do whatever extra tests :-).?

Thanks for all this work and getting this to completion!

/Dan

---
In reply to: [436438002](https://github.com/Azure/azure-uamqp-c/pull/271#issuecomment-436438002) [](ancestors = 436438002)

annatisch:
Thanks @dcristoloveanu - I'm sorry that you ended up having to fix my code!

I've just tried building from the anna-array-fix branch and am getting a number of these errors:
```
C:\Users\antisch\Documents\GitHub\azure-uamqp-python\src\vendor\azure-uamqp-c\src\amqpvalue.c(7145): error C2275: 'AMQP
_VALUE_DATA': illegal use of this type as an expression [C:\Users\antisch\Documents\GitHub\azure-uamqp-python\build\tem
p.win-amd64-3.6\Release\cmake\uamqp.vcxproj]
```
Am I missing some kind of configuration here?

annatisch:
Thanks @dcristoloveanu! Turns out I needed to update my azure-c-shared-utility dependency. All working now :)

PR #294, opened 2019-05-08
Fix building using installed dependencies

ewertons:


CMakeLists.txt
```diff
@@ -56,10 +56,8 @@ include("dependencies.cmake")
 include_directories(${MACRO_UTILS_INC_FOLDER})
 
 # Include the common build rules for C shared utility
-if(${use_installed_dependencies})
+if(NOT ${use_installed_dependencies})
```
massand: >NOT [](start = 3, length = 3)

Was this a bug? Or is this logic flipped from previous version due to specific reasons?

jenkins/linux_install_deps.sh
```diff
@@ -0,0 +1,36 @@
...
+echo "Build Root $build_root"
+
+sdk_build_folder=$build_root"/cmake/install_deps"
+
+# Build the SDK
+rm -rf $sdk_build_folder
+mkdir -p $sdk_build_folder
+pushd $sdk_build_folder
+cmake $build_root
+make install --jobs=$CORES
```
massand: >--jobs=$CORES [](start = 13, length = 13)

nit: does install require cores parameter? Can it be parallelized?

CMakeLists.txt
ewertons: This was a bug. Pasting here the conversation with @YoDaMa :

correct. It then means the code of those dependencies should not be expected to be where our submodules are set to be (in each project)

this was expecting the the code for the submodule was there, although the project is being built using installed dependencies

which is incorrect.

jenkins/linux_install_deps.sh
ewertons: make install builds if the project is not build, so thus the use of the parameter.
I do believe that $CORES will parallelize it.

PR #297, opened 2019-05-13
Fix building using installed dependencies

ewertons:


jebrando:
You didn't need to do ctest and the others?

PR #300, opened 2019-05-17
Additional fixes for vcpkg installation

ewertons:


build/.vsts-ci.yml
```diff
@@ -264,6 +264,41 @@ phases:
...
+  steps:
+  - script: |
+     if [ -f "jenkins/linux_install_deps.sh" ]
+     then
+      sudo chmod 755 jenkins/linux_install_deps.sh
+      sudo -E ./jenkins/linux_install_deps.sh
+      sudo jenkins/linux_install_deps.sh
+     fi
+    displayName: 'build'
+    env:
```
massand: Do we need these environment variables if we are not running the test?

ewertons: Probably not. Shall remove them in another PR so we can scan the whole yaml file for jobs to clean

CMakeLists.txt
```diff
@@ -50,10 +49,23 @@ if(NOT ${use_installed_dependencies})
...
         add_subdirectory(${CMAKE_CURRENT_LIST_DIR}/deps/azure-c-shared-utility)
     endif()
+else()
+    if (NOT azure_macro_utils_cFOUND)
+        find_package(azure_macro_utils_c REQUIRED CONFIG)
+    endif ()
+    if (NOT umock_cFOUND)
+        find_package(umock_c REQUIRED CONFIG)
+    endif ()
+    if (NOT azure_c_shared_utility_FOUND)
```
YoDaMa: why is this different from the rest of the variables of this format?

ewertons: you mean the trailing _ should be removed. Thanks, that's correct

PR #335, opened 2020-03-18
Check if pending_deliveries is not null before invoking singlylinkedlist_get_head_item

ewertons:
When the list is NULL, the logs keep printing "Invalid argument (list=NULL)" unnecessarily.


dcristoloveanu:
This is a duplicated PR of #337,

Issue #333, opened 2020-03-17
Session still accessing link after link_destroy invoked, resulting in SEGFAULT 


ewertons:
We are destroying a messagesender and its link afterwards, but right next the session still attempts to route an incoming message to such link, causing a memory access violation.

This is the stack of when the messagesender is destroyed:

Breakpoint 1, messagesender_destroy (message_sender=0x555555891a70) at /home/ewertons/azure-iot-sdk-c/uamqp/src/message_sender.c:751
751	    if (message_sender == NULL)
message_sender = 0x555555891a70
$20 = {link = 0x5555558983d0, message_count = 0, messages = 0x0, message_sender_state = MESSAGE_SENDER_STATE_OPEN, 
  on_message_sender_state_changed = 0x555555587690 <on_message_sender_state_changed_callback>, on_message_sender_state_changed_context = 0x5555558878a0, is_trace_on = 0}
#0  messagesender_destroy (message_sender=0x555555891a70) at /home/ewertons/azure-iot-sdk-c/uamqp/src/message_sender.c:751
#1  0x0000555555587636 in destroy_message_sender (instance=0x5555558878a0) at /home/ewertons/azure-iot-sdk-c/iothub_client/src/iothubtransport_amqp_messenger.c:614
#2  0x0000555555588dd0 in amqp_messenger_stop (messenger_handle=0x5555558878a0) at /home/ewertons/azure-iot-sdk-c/iothub_client/src/iothubtransport_amqp_messenger.c:1330
#3  0x0000555555585911 in twin_messenger_stop (twin_msgr_handle=0x5555558875a0) at /home/ewertons/azure-iot-sdk-c/iothub_client/src/iothubtransport_amqp_twin_messenger.c:2114
#4  0x0000555555576ad7 in amqp_device_stop (handle=0x555555887330) at /home/ewertons/azure-iot-sdk-c/iothub_client/src/iothubtransport_amqp_device.c:880
#5  0x0000555555571204 in IoTHubTransport_AMQP_Common_Device_DoWork (registered_device=0x555555887280) at /home/ewertons/azure-iot-sdk-c/iothub_client/src/iothubtransport_amqp_common.c:1185
#6  0x0000555555572358 in IoTHubTransport_AMQP_Common_DoWork (handle=0x5555558852c0) at /home/ewertons/azure-iot-sdk-c/iothub_client/src/iothubtransport_amqp_common.c:1638
[...]

PR #339, opened 2020-04-22
Fix for session_destroy_link_endpoint.

ewertons:
If an application calls link_destroy [void link_destroy(LINK_HANDLE link), link.c:872],

the link.c layer will then:
a. call session_destroy_link_endpoint [void session_destroy_link_endpoint(LINK_ENDPOINT_HANDLE link_endpoint), session.c:1162], AND
b. Free the link (!).

But session_destroy_link_endpoint will only remove the link endpoint if its state is different than LINK_ENDPOINT_STATE_ATTACHED [session.c:1174].

If then a message is received for the defunct link instance, session will try to route it anyway, which will result in a SEGFAULT.

Example of such scenario:
Device multiplexing where multiple links are established for all the features supposed by Azure IoT Hub.
If an error occurs during link attachment phase, app layer can decide to destroy all links of one of the devices IDs but not the session,
and a rogue message might still come from the server for a defunct link.

### DIFF
--- a/src/session.c
+++ b/src/session.c
@@ -1163,17 +1163,8 @@ void session_destroy_link_endpoint(LINK_ENDPOINT_HANDLE link_endpoint)
 {
     if (link_endpoint != NULL)
     {
-        LINK_ENDPOINT_INSTANCE* endpoint_instance = (LINK_ENDPOINT_INSTANCE*)link_endpoint;
-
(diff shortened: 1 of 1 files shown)


ewertons:
Related to https://github.com/Azure/azure-uamqp-c/issues/333

ewertons:
This actually is not the correct fix for the uamqp design. I will post another fix in the proper location.

PR #341, opened 2020-05-12
Avoid propagating FLOW for detaching links

ewertons:
Addresses https://github.com/Azure/azure-uamqp-c/issues/333

### DIFF
--- a/src/session.c
+++ b/src/session.c
@@ -627,7 +627,8 @@ static void on_frame_received(void* context, AMQP_VALUE performative, uint32_t p
                 while ((session_instance->remote_incoming_window > 0) && (i < session_instance->link_endpoint_count))
                 {
                     /* notify the caller that it can send here */
-                    if (session_instance->link_endpoints[i]->on_session_flow_on != NULL)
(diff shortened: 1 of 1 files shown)


src/session.c
```diff
@@ -627,7 +627,8 @@ static void on_frame_received(void* context, AMQP_VALUE performative, uint32_t p
                 while ((session_instance->remote_incoming_window > 0) && (i < session_instance->link_endpoint_count))
                 {
                     /* notify the caller that it can send here */
-                    if (session_instance->link_endpoints[i]->on_session_flow_on != NULL)
+                    if (session_instance->link_endpoints[i]->link_endpoint_state != LINK_ENDPOINT_STATE_DETACHING &&
```
danewalton, line 630: Completely ignorant question here....would it be better to have this be:
```suggestion
                    if (session_instance->link_endpoints[i]->link_endpoint_state == LINK_ENDPOINT_STATE_ATTACHED &&
```

ewertons: Sure, no problem. Thanks for the question. That's probably a good option and would be in sync with the uamqp spec, but I would not want to make that change now since it would be too restrictive and probably prone to unearthing other possible bugs.

PR #346, opened 2020-08-11
Update do while loop to do away with compiler warning implementation

danewalton:


danewalton:
This can't be merged since this file is autogenerated. Will update the autogeneration file.

PR #347, opened 2020-08-12
remove do/while compiler warning workaround

danewalton:


src/amqp_definitions.c
```diff
@@ -7,7 +7,7 @@
 // The generator that produces it is located at /uamqp_generator/uamqp_generator.sln
 
 #include <stdlib.h>
-#include "azure_macro_utils/macro_utils.h"
```
ewertons, line 10: I think this file uses macro utils everywhere. It should have this header included, not sure why the autogenerator excluded it. Could you readd it manually please?

PR #357, opened 2020-12-04
Fix potential double free of link endpoint by link.c

ewertons:


src/session.c
```diff
@@ -104,6 +106,11 @@ static void remove_link_endpoint(LINK_ENDPOINT_HANDLE link_endpoint)
 
 static void free_link_endpoint(LINK_ENDPOINT_HANDLE link_endpoint)
 {
+    if (link_endpoint->on_link_endpoint_destroyed_callback != NULL)
+    {
```
momuno, line 110: So, if the on_link_ep_destr_cb is set, and the link_endpoint and link (aka context) check-out (via on_link_ep_destr), then is it safe to assume that the link_endpoint has already been freed by the session?  And assume if on_link_sp_dest_cb has not been set, or link_endpoint and link do not check-out, then the link_endpoint has NOT been freed by the session? 

It's not clear to me to make this assumption (other than your comment in link.c line 712 implies I should).  Perhaps a few more comments explaining how this relationship works could help a future reader understand this.  I'm concerned a future change elsewhere in code could introduce a memory leak if the coder does not fully understand.  What do you think?

inc/azure_uamqp_c/session.h
```diff
@@ -58,10 +58,12 @@ MU_DEFINE_ENUM(SESSION_STATE, SESSION_STATE_VALUES)
 MU_DEFINE_ENUM(SESSION_SEND_TRANSFER_RESULT, SESSION_SEND_TRANSFER_RESULT_VALUES)
 
     typedef void(*LINK_ENDPOINT_FRAME_RECEIVED_CALLBACK)(void* context, AMQP_VALUE performative, uint32_t frame_payload_size, const unsigned char* payload_bytes);
+    typedef void(*ON_LINK_ENDPOINT_DESTROYED_CALLBACK)(LINK_ENDPOINT_HANDLE handle, void* context);
     typedef void(*ON_SESSION_STATE_CHANGED)(void* context, SESSION_STATE new_session_state, SESSION_STATE previous_session_state);
     typedef void(*ON_SESSION_FLOW_ON)(void* context);
     typedef bool(*ON_LINK_ATTACHED)(void* context, LINK_ENDPOINT_HANDLE new_link_endpoint, const char* name, role role, AMQP_VALUE source, AMQP_VALUE target, fields properties);
 
+
```
momuno, line 66: nit: extra line remove

src/session.c
ewertons: Makes sense. I'll add the note explaining why this is needed.

ericwol-msft (approved):

inc/azure_uamqp_c/session.h
ewertons: Fixed.

PR #361, opened 2020-12-15
link desired capabilities support

yunhaoling:
According to the [amqp spec](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html) section 2.7.3 Attach -- Attach frame should support desired-capabilities:

image

This PR is to address the missing attach desired-capabilities support.

### DIFF
--- a/src/link.c
+++ b/src/link.c
@@ -61,6 +61,7 @@ typedef struct LINK_INSTANCE_TAG
     uint32_t max_link_credit;
     uint32_t available;
     fields attach_properties;
+    AMQP_VALUE desired_capabilities;
     bool is_underlying_session_begun;
     bool is_closed;
     unsigned char* received_payload;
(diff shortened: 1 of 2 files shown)


src/link.c
```diff
@@ -1108,6 +1120,59 @@ int link_get_peer_max_message_size(LINK_HANDLE link, uint64_t* peer_max_message_
...
+        }
+    }
+    return result;
+}
+
+int link_set_desired_capabilities(LINK_HANDLE link, AMQP_VALUE desired_capabilities)
+{
+    int result;
+
+    if (link == NULL)
```
dcristoloveanu, line 1154: >if (link == NULL) [](http://example.com/codeflow?start=4&length=17)

There should probably be a check for desired_capabilities here also.

src/link.c
```diff
@@ -1108,6 +1120,59 @@ int link_get_peer_max_message_size(LINK_HANDLE link, uint64_t* peer_max_message_
...
+    {
+        LogError("NULL link");
```
dcristoloveanu, line 1156: >rror("NULL link"); [](http://example.com/codeflow?start=12&length=18)

Consider logging desired_capabilities also.

dcristoloveanu:
Sorry for the long delay. I have not visited this repo in a bit :-).
Besides the 2 minor comments LGTM!

dcristoloveanu (approved):
:shipit:

PR #363, opened 2020-12-16
[Discussion] Proactive vs Passive Flow Control

yunhaoling:
This PR is for discussion on proactive link vs passive link flow control.

Current implementation
- new flow is sent via link_frame_received when `current_link_credit == 0`.
  - my understanding is that the link_frame_received is called whenever there's a frame target at the link
  - link doesn't have the actual control over the send_flow in this implementation, the send_flow is triggered by connection receiving transfer frames.

This PR propose the send flow control moves from `link_frame_received` to `link_dowork` which is controlled by the upper layer.

thoughts?

### DIFF
--- a/src/link.c
+++ b/src/link.c
@@ -57,7 +57,7 @@ typedef struct LINK_INSTANCE_TAG
     sequence_no initial_delivery_count;
     uint64_t max_message_size;
     uint64_t peer_max_message_size;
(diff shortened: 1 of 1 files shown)


ewertons:
This is way outdated. Apologies for not handling it earlier.

PR #364, opened 2020-12-16
[Pending] Fixed seg fault in tearing down a failed link with unsent pending messages

yunhaoling:
https://github.com/Azure/azure-uamqp-python/pull/140

This is a fix for seg fault in azure-uamqp-python. However I'm currently clueless on how it fixes a seg fault (in what case/how the original code could trigger a seg fault). Will bring more context after I dig deeper also wondering if I could get some eyes.


alseambusher:
@yunhaoling can you merge this?

PR #376, opened 2021-01-27
AMQP fuzzing fixes

ericwol-msft:


src/amqpvalue.c
```diff
@@ -12,6 +12,10 @@
 #include "azure_uamqp_c/amqpvalue.h"
 #include "azure_c_shared_utility/refcount.h"
 
+// max alloc size 100MB
```
ericwol-msft, line 15: note the max hub payload size if 64k, so this limit should be fine

src/amqpvalue.c
```diff
@@ -4959,6 +4969,7 @@ static int internal_decoder_decode_bytes(INTERNAL_DECODER_DATA* internal_decoder
                 {
                     AMQP_VALUE_DATA* descriptor;
                     internal_decoder_data->decode_to_value->type = AMQP_TYPE_DESCRIBED;
+                    internal_decoder_data->decode_to_value->value.described_value.value = NULL;
```
ewertons, line 4972: If this is not NULL already, its value should be REF COUNT DEC before nulling it out, otherwise we might have a memory leak. Please check where it gets set, on 
https://github.com/Azure/azure-uamqp-c/pull/376/files#diff-7ec5b31fa5071f1609135ee9c21730636a478036d89115c51781e7353fdcbaf0R5435

ewertons (approved):
Please see comments.

danewalton (approved):

src/amqpvalue.c
ericwol-msft: moved zeroing out the struct where it's allocated.

src/frame_codec.c
```diff
@@ -302,7 +304,8 @@ int frame_codec_receive_bytes(FRAME_CODEC_HANDLE frame_codec, const unsigned cha
                         frame_codec_data->receive_frame_pos = 0;
 
                         /* Codes_SRS_FRAME_CODEC_01_102: [frame_codec_receive_bytes shall allocate memory to hold the frame_body bytes.] */
-                        frame_codec_data->receive_frame_bytes = (unsigned char*)malloc(frame_codec_data->receive_frame_size - 6);
+                        frame_codec_data->receive_frame_malloc_size = frame_codec_data->receive_frame_size - 6;
```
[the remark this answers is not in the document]
ericwol-msft: out of scope

PR #384, opened 2021-03-03
bug 9443347: [FuzzAMQP] heap-use-after-free link_frame_received()

ericwol-msft:
[FuzzAMQP] heap-use-after-free link_frame_received() @ azure-iot-sdk-c/uamqp/src/link.c:330


ewertons (approved):
It makes sense as it is written. I hope nothing will break.

PR #400, opened 2021-06-02
Update submodules to remove 'www' prefix

laumann:
When cloning this repo as a dependency of [0] and updating submodules,
Git displays warnings like this:

    warning: redirecting to https://github.com/Azure/azure-ctest.git/

Fix by removing the "www" prefix from the module URLs.

[0] https://github.com/Azure/azure-iot-sdk-c


laumann:
@ewertons Not sure if I should ping anyone, so here goes :-)

momuno (approved):

PR #414, opened 2021-12-03
Fixes for SDLNativeRules issues

ericwol-msft:


src/sasl_plain.c
```diff
@@ -66,8 +67,17 @@ CONCRETE_SASL_MECHANISM_HANDLE saslplain_create(void* config)
                 else
                 {
                     /* Ignore UTF8 for now */
-                    result->init_bytes = (unsigned char*)malloc(authzid_length + authcid_length + passwd_length + 2);
-                    if (result->init_bytes == NULL)
+                    //size_t malloc_size = authzid_length + authcid_length + passwd_length + 2);  // use safe add
```
ewertons, line 70: Remove?

src/sasl_plain.c
```diff
@@ -66,8 +67,17 @@ CONCRETE_SASL_MECHANISM_HANDLE saslplain_create(void* config)
...
+                    size_t malloc_size = safe_add_size_t(authzid_length, authcid_length);
+                    malloc_size = safe_add_size_t(malloc_size, passwd_length);
+                    malloc_size = safe_add_size_t(malloc_size, 2);
+
+                    if (malloc_size == SIZE_MAX)
```
ewertons, line 75: I have similar comments as in c-util about disregarding SIZE_MAX as a valid size, but will refrain from posting all over because we are intentionally choosing to do so.

src/session.c
```diff
@@ -1607,9 +1608,13 @@ SESSION_SEND_TRANSFER_RESULT session_send_transfer(LINK_ENDPOINT_HANDLE link_end
                                             }
                                         }
 
-                                        transfer_frame_payload_count = (uint32_t)(temp_current_payload_index - current_payload_index + 1);
-                                        transfer_frame_payloads = (PAYLOAD*)calloc(1, (transfer_frame_payload_count * sizeof(PAYLOAD)));
-                                        if (transfer_frame_payloads == NULL)
+                                        //transfer_frame_payload_len = (uint32_t)(temp_current_payload_index - current_payload_index + 1); // use safe int
```
ewertons, line 1611: Remove?

src/session.c
```diff
@@ -1607,9 +1608,13 @@ SESSION_SEND_TRANSFER_RESULT session_send_transfer(LINK_ENDPOINT_HANDLE link_end
...
+                                        size_t size = safe_subtract_size_t(temp_current_payload_index, current_payload_index);
```
ewertons, line 1612: Could we have a more contextualized name?

ewertons (approved):
Approved with comments. Please take a look on the dead code (commented lines) and remove. Thanks Eric!

src/sasl_plain.c
ericwol-msft: no, I want it there to show the expected math

src/session.c
ericwol-msft: no, I want it there to show the expected math

src/sasl_plain.c
ericwol-msft: As discussed yesterday. safe_add is a clamping function. mallocing all memory is an error condition.

PR #426, opened 2022-09-20
Rename data definition name to fix OACR unambiguous symbol

savukyam:
Hitting into OACR static analysis build issue on typedef for amqp_binary in https://github.com/Azure/azure-uamqp-c/blob/f7acb44aa8f61d9958ed2786831a53fd0a4a566e/inc/azure_uamqp_c/amqp_definitions_data.h

This PR tries to fix the same by renaming the type name.

### DIFF
--- a/uamqp_generator/amqp_definitions.xml
+++ b/uamqp_generator/amqp_definitions.xml
@@ -246,7 +246,7 @@
     <type name="application-properties" class="restricted" source="map" provides="section">
       <descriptor name="amqp:application-properties:map" code="0x00000000:0x00000074"/>
     </type>
-    <type name="data" class="restricted" source="binary" provides="section">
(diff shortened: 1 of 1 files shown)


ewertons:
Superseded by #485, merged as 91efcbd.

This PR changed only `uamqp_generator/amqp_definitions.xml`. That file is the generator input; the generator is not run by the build, so the change did not affect any shipped header or source. Regenerating from it would also have renamed the published header file, `amqpvalue_create_data`, `is_data_type_by_descriptor` and the `amqpvalue_get_data` macro, which breaks existing consumers.

#485 fixes the reported ambiguity by renaming only the typedef to `amqp_data` and keeping `data` as a backwards-compatible alias guarded by `AMQP_DEFINITIONS_NO_LEGACY_DATA_TYPE`, with the change made in the generator and the regenerated output included. Thanks for the report.

PR #430, opened 2023-02-17
Fixed race condition in sendAsync related to messages; fixed local_server_sample to no longer AV.

LarryOsterman:
Fixes a race condition in `remove_pending_message_by_index`.

The existing code in remove_pending_message_by_index had:

```cpp
        free(message_sender->messages);
        message_sender->messages = NULL;
```

The problem is that if there was a context switch between the call to `free()` and before the `message_sender->messages` was set to NULL, then the call to reallocate the messages array in line 917:
```cpp
               ASYNC_OPERATION_HANDLE* new_messages = (ASYNC_OPERATION_HANDLE*)realloc(message_sender->messages, sizeof(ASYNC_OPERATION_HANDLE) * (message_sender->message_count + 1));
 
```
would reference freed memory and (if you were lucky) AV. If you weren't lucky, it corrupted memory.

### DIFF
--- a/samples/local_server_sample/main.c
+++ b/samples/local_server_sample/main.c
@@ -18,117 +18,127 @@ static MESSAGE_RECEIVER_HANDLE message_receiver;
 
 static void on_message_receiver_state_changed(const void* context, MESSAGE_RECEIVER_STATE new_state, MESSAGE_RECEIVER_STATE previous_state)
 {
-    (void)context, new_state, previous_state;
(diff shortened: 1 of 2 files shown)


LarryOsterman:
One other note: This only reduces the window associated with the bug - there is other code which checks the `message_count` field and decides to reallocate/free the array based on that field value, this also can be subject to race conditions.

PR #444, opened 2023-12-05
int overflow for binary_value size malloc

ericwol-msft:


src/amqpvalue.c
```diff
@@ -5912,7 +5912,16 @@ static int internal_decoder_decode_bytes(INTERNAL_DECODER_DATA* internal_decoder
                             }
                             else
                             {
-                                internal_decoder_data->decode_to_value->value.binary_value.bytes = (unsigned char*)malloc((size_t)internal_decoder_data->decode_to_value->value.binary_value.length + 1);
+                                size_t malloc_size = (size_t)internal_decoder_data->decode_to_value->value.binary_value.length + 1;
+                                if (malloc_size == 0)
```
jspaith, line 5916: Nit I'd consider logging an error here on overflow.  Since the `MU_FAILURE` doesn't mean anything letting the error be handled a few lines below in the `bytes == NULL` is fine.

src/amqpvalue.c
```diff
@@ -5912,7 +5912,16 @@ static int internal_decoder_decode_bytes(INTERNAL_DECODER_DATA* internal_decoder
...
+                                {
+                                    internal_decoder_data->decode_to_value->value.binary_value.bytes = NULL;
+                                }
+                                else
+                                {
+                                    internal_decoder_data->decode_to_value->value.binary_value.bytes = (unsigned char*)malloc((size_t)internal_decoder_data->decode_to_value->value.binary_value.length + 1);
```
jspaith, line 5922: Nit - you've calculated `malloc_size` already you may as well use it here.

jspaith (approved):

src/amqpvalue.c
vaavva, line 5916: Is this possible to be 0 since it's a size_t + 1?

vaavva (approved):

PR #446, opened 2024-01-09
Added int overflow checks for amqp value

ericwol-msft:


ewertons:
amqpvalue.c is not auto-generated, which is good for this PR.
The one that is auto-gen is amqp_definitions.c
https://github.com/Azure/azure-uamqp-c/blob/master/src/amqp_definitions.c#L6

src/amqpvalue.c
```diff
@@ -1069,8 +1069,18 @@ AMQP_VALUE amqpvalue_create_string(const char* value)
     else
     {
         size_t length = strlen(value);
+        size_t malloc_size = length + 1;
+
+        if (malloc_size == 0)
```
ewertons, line 1074:
```suggestion
        // If the result of malloc_size is zero it means it had a type overflow (size_t is an unsigned type).
        // It is very unlikely but could happen.
        if (malloc_size == 0)
```

src/amqpvalue.c
```diff
@@ -5975,7 +5985,17 @@ static int internal_decoder_decode_bytes(INTERNAL_DECODER_DATA* internal_decoder
                         buffer++;
                         size--;
 
-                        internal_decoder_data->decode_to_value->value.string_value.chars = (char*)malloc((size_t)internal_decoder_data->decode_value_state.string_value_state.length + 1);
+                        size_t malloc_size = (size_t)internal_decoder_data->decode_value_state.string_value_state.length + 1;
+                        if (malloc_size == 0)
```
ewertons, line 5989:
```suggestion
                        // If the result of malloc_size is zero it means it had a type overflow (size_t is an unsigned type).
                        // It is very unlikely but could happen.
                        if (malloc_size == 0)
```

src/amqpvalue.c
```diff
@@ -6038,7 +6058,17 @@ static int internal_decoder_decode_bytes(INTERNAL_DECODER_DATA* internal_decoder
 
                         if (internal_decoder_data->bytes_decoded == 4)
                         {
-                            internal_decoder_data->decode_to_value->value.string_value.chars = (char*)malloc((size_t)internal_decoder_data->decode_value_state.string_value_state.length + 1);
+                            size_t malloc_size = (size_t)internal_decoder_data->decode_value_state.string_value_state.length + 1;
+                            if (malloc_size == 0)
```
ewertons, line 6062:
```suggestion
                            // If the result of malloc_size is zero it means it had a type overflow (size_t is an unsigned type).
                            // It is very unlikely but could happen.
                            if (malloc_size == 0)
```

src/amqpvalue.c
```diff
@@ -6104,7 +6134,17 @@ static int internal_decoder_decode_bytes(INTERNAL_DECODER_DATA* internal_decoder
                         buffer++;
                         size--;
 
-                        internal_decoder_data->decode_to_value->value.symbol_value.chars = (char*)malloc((size_t)internal_decoder_data->decode_value_state.symbol_value_state.length + 1);
+                        size_t malloc_size = (size_t)internal_decoder_data->decode_value_state.symbol_value_state.length + 1;
+                        if (malloc_size == 0)
```
ewertons, line 6138:
```suggestion
                        // If the result of malloc_size is zero it means it had a type overflow (size_t is an unsigned type).
                        // It is very unlikely but could happen.
                        if (malloc_size == 0)
```

src/amqpvalue.c
```diff
@@ -6167,7 +6207,17 @@ static int internal_decoder_decode_bytes(INTERNAL_DECODER_DATA* internal_decoder
 
                         if (internal_decoder_data->bytes_decoded == 4)
                         {
-                            internal_decoder_data->decode_to_value->value.symbol_value.chars = (char*)malloc((size_t)internal_decoder_data->decode_value_state.symbol_value_state.length + 1);
+                            size_t malloc_size = (size_t)internal_decoder_data->decode_value_state.symbol_value_state.length + 1;
+                            if (malloc_size == 0)
```
ewertons, line 6211:
```suggestion
                            // If the result of malloc_size is zero it means it had a type overflow (size_t is an unsigned type).
                            // It is very unlikely but could happen.
                            if (malloc_size == 0)
```

src/amqpvalue.c
```diff
@@ -6511,8 +6561,19 @@ static int internal_decoder_decode_bytes(INTERNAL_DECODER_DATA* internal_decoder
                                 uint32_t i;
 
                                 internal_decoder_data->decode_to_value->value.map_value.pair_count /= 2;
+                                size_t malloc_size = safe_multiply_size_t(sizeof(AMQP_MAP_KEY_VALUE_PAIR), ((size_t)internal_decoder_data->decode_to_value->value.map_value.pair_count);
```
ewertons, line 6564: Excellent!
https://github.com/Azure/azure-c-shared-utility/blob/master/inc/azure_c_shared_utility/safe_math.h#L15

src/amqpvalue.c
```diff
@@ -6550,13 +6611,21 @@ static int internal_decoder_decode_bytes(INTERNAL_DECODER_DATA* internal_decoder
                                     uint32_t i;
 
                                     internal_decoder_data->decode_to_value->value.map_value.pair_count /= 2;
+                                    size_t malloc_size = safe_multiply_size_t((size_t)internal_decoder_data->decode_to_value->value.map_value.pair_count, 2);
```
ewertons, line 6614: Excellent!
https://github.com/Azure/azure-c-shared-utility/blob/master/inc/azure_c_shared_utility/safe_math.h#L15

src/amqpvalue.c
jspaith, line 6564: Nit - not sure why we divide the pair_count by two on first line of this block and then `malloc_size = safe_multiply_size_t(malloc_size, 2);` a few lines later?  This is pre-existing code so maybe outside scope of this change though.

src/amqpvalue.c
```diff
@@ -1069,8 +1069,18 @@ AMQP_VALUE amqpvalue_create_string(const char* value)
     else
```
ewertons, line 1069: Another place where there could be overflow as well:

In this block it should be checked the way you implemented in your PR.
The block assumes size_t == uint32_t, which is incorrect.
https://github.com/Azure/azure-uamqp-c/blob/28f603aaa0fa36b26e0eb4420dea1a9ef1cb3719/src/amqpvalue.c#L1177

There were no other locations I found that could have a type overflow followed by malloc.

ewertons (approved):
Approved with requests for change.

src/amqpvalue.c
```diff
@@ -6550,13 +6611,21 @@ static int internal_decoder_decode_bytes(INTERNAL_DECODER_DATA* internal_decoder
...
                                     internal_decoder_data->decode_to_value->value.map_value.pair_count /= 2;
+                                    size_t malloc_size = safe_multiply_size_t((size_t)internal_decoder_data->decode_to_value->value.map_value.pair_count, 2);
+                                    malloc_size = safe_multiply_size_t(sizeof(AMQP_MAP_KEY_VALUE_PAIR), malloc_size);
+
                                     if (internal_decoder_data->decode_to_value->value.map_value.pair_count > MAX_AMQPVALUE_ITEM_COUNT)
                                     {
                                         LogError("AMQP list map count exceeded MAX_AMQPVALUE_ITEM_COUNT");
                                         result = MU_FAILURE;
                                     }
+                                    else if (malloc_size == SIZE_MAX)
```
jspaith, line 6622: Why do we check against `SIZE_MAX` here but not other places?  Shouldn't the `safe_multiple_size_t` protect us from overflow and then on off-shot we ended up with `SIZE_MAX` we'd rely on the malloc() call to fail but otherwise be OK knowing we ourselves hadn't over-flown?

jspaith (approved):

src/amqpvalue.c
ericwol-msft: should I change the **IF** to

`if (length > UINT32_MAX)`

src/amqpvalue.c
ericwol-msft: It's a known error state, so probably should not call malloc, also it gives more useful trace statement.

PR #455, opened 2024-02-07
Fix potential null pointer dereference

szsam:
Move dereference of message_with_callback inside the if block that checks the pointer is non-null.


ewertons (approved):
Very nice. Thanks for this fix!

ewertons:
Excellent, @szsam , thank you for updating the branch!

PR #480, opened 2026-07-28
Harden the decoder against malformed input

cataggar:
## The bug

`decodeListItems`, `decodeMapEntries`, and `decodeArrayItems` each allocate an uninitialized array and then, on any error, run an errdefer that calls `deinit` on **every** element:

```zig
const entries = try allocator.alloc(MapEntry, pair_count);
errdefer {
    for (entries) |*entry| {          // <- all of them, not just the decoded ones
        @constCast(&entry.key).deinit(allocator);
        @constCast(&entry.value).deinit(allocator);
    }
    allocator.free(entries);
}
```

Only the elements decoded so far are initialized. The cleanup walks whatever the allocator happened to leave in the tail and frees it as if it were an `AmqpValue`.

The count and the element bytes both come straight off the wire, so any peer can reach this. Three bytes are enough:

```zig
decode(allocator, &.{ 0xc1, 0x02, 0x02 })  // map8, one pair declared, no body
```

In a safe build that panics with `switch on corrupt value`. In a release build without safety checks it frees arbitrary pointers.

I hit this while decoding AMQP frames in `azure-sdk-for-zig`, which reaches `decode` on every inbound frame body.

## Second issue: unbounded allocation

The element count is read from the wire and used to size an allocation with no cross-check against the data actually present — `size` is read and then explicitly discarded (`_ = size;`). A nine-byte frame claiming `0xffffffff` list items asks for 64 GiB:

```zig
decode(allocator, &.{ 0xd0, 0x00, 0x00, 0x00, 0x04, 0xff, 0xff, 0xff, 0xff })
```

## The fix

Each of the three now tracks how many elements it has written and cleans up only that prefix. `decodeMapEntries` additionally frees the key it just decoded when the matching value fails, which previously leaked.

For the allocation bound: every element needs at least one byte and every map pair at least two, so the count is rejected against the remaining length before allocating. This is a cheap lower bound, not full `size` validation — it is enough to stop the count alone from driving the allocation.

## Testing

48 -> 55 tests, covering truncation at each of the three levels, a nested truncated list, partial map cleanup where the key decodes but the value does not, and the oversized counts.

Verified the tests actually catch the original bug: reverting just the prefix tracking (`entries[0..filled]` back to `entries`) crashes two of them with the original panic.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,8 @@
 
 /.vscode/*
 /**/.vs/*
+
+# Zig
+zig-out/
+zig-cache/
+.zig-cache/
--- a/build.zig
+++ b/build.zig
@@ -1,0 +1,54 @@
+const std = @import("std");
+
+pub fn build(b: *std.Build) void {
+    const target = b.standardTargetOptions(.{});
+    const optimize = b.standardOptimizeOption(.{});
+
+    // Main library module
+    const lib_mod = b.createModule(.{
+        .root_source_file = b.path("src/zig/uamqp.zig"),
+        .target = target,
+        .optimize = optimize,
+    });
+
+    // Static library artifact
+    const lib = b.addLibrary(.{
+        .linkage = .static,
+        .name = "uamqp",
+        .root_module = lib_mod,
+    });
+    b.installArtifact(lib);
+
+    // Unit tests
+    const test_mod = b.createModule(.{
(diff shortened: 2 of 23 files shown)


cataggar:
Opened against the wrong repository by mistake; this targets the Zig port in a fork, not azure-uamqp-c. Closing. Apologies for the noise.

